PR #2, opened 2018-05-24
Minor improvements

niklasad1:
Some minor improvements for performance and readability!

I haven't tested the `s/write!/writeln!` but I think the code becomes easier to read!


src/header.rs
```diff
@@ -54,56 +51,29 @@ struct TbfHeaderWriteableFlashRegion {
...
-            self.total_size,
-            self.total_size,
-            self.flags,
-            self.flags,
-        )
+        writeln!(f)?;
+        writeln!(f, "version: {:>8} {:>#10X}", self.version,self.version)?;
+        writeln!(f, "header_size: {:>8} {:>#10X}", self.header_size, self.header_size)?;
+        writeln!(f, "total_size: {:>8} {:>#10X}", self.total_size, self.total_size)?;
+        writeln!(f, "flags: {:>8} {:>#10X}", self.flags, self.flags)
```
bradjc: This removes all of the alignment that Pat added to make the printout easier to read.

src/header.rs
```diff
@@ -206,7 +176,7 @@ impl TbfHeader {
         }
 
         // Return the length by generating the header and seeing how long it is.
-        self.generate().unwrap().get_ref().len()
+        self.generate().unwrap_or_else(|_| io::Cursor::new(Vec::new())).get_ref().len()
```
bradjc: This will make this line never fail, correct? Why would we want that? If the generate call cannot be unwrapped, there is no point in returning a header length of 0.

src/header.rs
niklasad1: Ok, I can revert this! I guess this applies to all `write! stuff`?

src/header.rs
niklasad1: Yeah, that is true in general I'm allergic to `raw unwrap()`!

I didn't think this through, ok If I change to `expect("Error: header is empty")`

niklasad1:
ping @bradjc @alevy 

merge? :P

PR #5, opened 2018-11-25
Improve cli interface and user experience

Nicoretti:
1. Remove "required" options

    elf2tab have had required options, these generaly should be avoided,
    because they have various usability drawbacks:

    * It forces the user not only to specify the values, but also
          to specify the option long/short - name
    * The usage / help does not make it immediatly clear they they are
          required (a user generally expects an option to be optional)

    Therefore "required options" should be replaced either with either:

    * options which have a reasonable default
    * or with required arguments

2. Add color

    Color will make it easier to read.

3. Ensure at lest one application elf file has to be specified

    The previous implementation and usage of the cli suggested that the
    ELF paramater(s) are completely optional.

4. Add basic tests for required parameters

5. Bump version number

    Increase patch version number so its clear there have been some
    changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,13 @@ Usage
 
 ```
 USAGE:
-    elf2tab [FLAGS] [OPTIONS] --app-heap <APP_HEAP_SIZE> --kernel-heap <KERNEL_HEAP_SIZE> --stack <STACK_SIZE> -o <TAB> [ELF]...
+    elf2tab [FLAGS] [--protected-region-size=<protected-region-size>]
+                    [--package-name=<pkg-name>] [--output-file=<filename>] <elf>...
+    elf2tab [FLAGS] [--proteced-region-size=<protected-region-size>] [--package-name=<pkg-name>]
+                    [--output-file=<filename>] [--minimum-ram-size=<min-ram-size>] <elf>...
+    elf2tab [FLAGS] [--proteced-region-size=<protected-region-size>]
+                    [--package-name=<pkg-name>] [--output-file=<filename>]
+                    [--app-heap=<heap-size>] [--kernel-heap=<kernel-heap-size>] [--stack=<stack-size>] <elf>...
 
 FLAGS:
     -h, --help       Prints help information
@@ -19,14 +25,16 @@ FLAGS:
     -v, --verbose    Be verbose
 
 OPTIONS:
-        --app-heap <APP_HEAP_SIZE>          App heap size in bytes
-        --kernel-heap <KERNEL_HEAP_SIZE>    Kernel heap size in bytes
-    -n <PACKAGE_NAME>                       Package Name
-        --stack <STACK_SIZE>                Stack size in bytes
-    -o <TAB>                                Output file name
+    -o, --output-file <filename>             output file name [default: TockApp.tab]
+        --app-heap <heap-size>               in bytes [default: 1024]
+        --kernel-heap <kernel-heap-size>     in bytes [default: 1024]
+        --minimum-ram-size <min-ram-size>    in bytes
+    -p, --package-name <pkg-name>            package name
+        --protected-region-size <protected-region-size>    Size of the protected region (including headers)
+        --stack <stack-size>                 in bytes [default: 2048]
 
 ARGS:
-    <ELF>...    App elf files
+    <elf>...    application file(s) to package
 ```
 
 For example, converting a "blink" app from a compiled .elf file (for a Cortex-M4
@@ -37,7 +45,7 @@ device) with this tool would look like:
 It also supports (and encourages!) combing .elf files for multiple architectures
 into a single tab:
 
-	$ elf2tab -o blink.tab -n blink --stack 1024 --app-heap 1024 --kernel-heap 1024 cortex-m0.elf cortex-m3.elf cortex-m4.elf
+    $ elf2tab -o blink.tab -n blink --stack 1024 --app-heap 1024 --kernel-heap 1024 cortex-m0.elf cortex-m3.elf cortex-m4.elf
 
 
 Compiling elf2tab
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -2,10 +2,13 @@ use std::path::PathBuf;
 use structopt;
 
 fn usage() -> &'static str {
-    "elf2tab [FLAGS] [--package-name=<pkg-name>] [--output-file=<filename>] <elf>...
-    elf2tab [FLAGS] [--package-name=<pkg-name>] [--output-file=<filename>] [--minimum-ram-size=<min-ram-size>] <elf>...
-    elf2tab [FLAGS] [--package-name=<pkg-name>] [--output-file=<filename>] [--app-heap=<heap-size>]
-                    [--kernel-heap=<kernel-heap-size>] [--stack=<stack-size>] <elf>..."
+    "elf2tab [FLAGS] [--protected-region-size=<protected-region-size>]
+                    [--package-name=<pkg-name>] [--output-file=<filename>] <elf>...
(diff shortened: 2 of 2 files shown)


Nicoretti:
Hi @bradjc,

I just come to use the elf2tab and I think the cli interface can be improved slightly so it will be more clear and easier to the cli user. If I missed something or you want the too adjust something just let me know (happy to get some feedback and/or to be engaged into a technical discussion).

greez
Nico

bradjc:
Hi @Nicoretti 

 >   elf2tab have had required options, these generaly should be avoided,
    because they have various usability drawbacks:

I bet we can come up with reasonable defaults for all of the options.

>    Add color

Does `elf2tab` even print anything? In any case, given the tool will almost always be run as part of a build script, I don't see the effort for getting color to work to be worth it. `tockloader` on the other hand...

>    Ensure at least one application elf file has to be specified

This should be happening, and if it isn't it's probably due to an unfamiliarity with command line parsing in Rust.

>    Add basic tests for required parameters

PRs welcome :)

>    Bump version number

Yep, and I imagine this would be required for crates.io anyway.

Nicoretti:
Hi Brad,

Thx for your feedback.

> I bet we can come up with reasonable defaults for all of the options.
I can see why this is and should be the preffered solution, by solving it this way we do
not break existing usages of the tool in existing scripts etc. and it still will be easier to use.

Do you have any suggestions on what the defaults should be? otherwise I would orient myself
on the usages in libtock-rs examples  `--stack 2048 --app-heap 1024 --kernel-heap 1024`
We also could use a default for the a output flag, if so I would use `-o app.tab`

> Does elf2tab even print anything? In any case, given the tool will almost always be run as part of a build script, I don't see the effort for getting color to work to be worth it. tockloader on the other hand...

This color is just for the cli help and comes without any extra effort, it is just an extra setting for clap , which is already part of the project. 
If you have concerns because of potentional parsing/automation, do not worry, clap checks where the output is streamed to (is_tty), therefore e.g. `user@host ~ elf2tab --help > out.txt` will contain no color escape sequences.

cli

cli


> Ensure at least one application elf file has to be specified
> 
> This should be happening, and if it isn't it's probably due to an unfamiliarity with command line parsing in Rust.

Not specifying an ELF file worked perfectly fine with the old cli (see cli and test run screenshots)

Old CLI
old_cli

Test run
no_elf

If you are ok with the defaults or have better alternatives I'll update the PR accordingly.

BTW
I also saw `-o` only has a short option, I would add the additional long version of it too.
-> It is a good practise to do so

best regards
Nico

alevy:
The colors are great! And I agree on not having required arguments.

However, I think my vote would be for have two possible sets of arguments for specifying memory, as well as a default, as follows:

  * `--stack`, `--app-heap`, `--kernel-heap` as they are currently used
  * or, `--minimum_ram_size`, which is used directly in place of the current computation
  * or, default to just summing the memory in each of the PT_LOAD sections, as is currently done (the ELF can easily stack and heap space explicitly in the BSS section itself).

The third is effectively equivalent to defaulting each of the first three options to `0` and making them each independently optional, so that's also fine.

Nicoretti:
Hi Amit,

thx for your feedback, I'll try to update the PR accordingly even though I am not 100% if we can do exactly what you wanted using clap/structop.

Remarks:
* if we introduce the options `--minimum_ram_size` we should name it `--minimum-ram-size`
  (It is good and common practise to use `-` between words in long options)

* If I understand you correctly you proposing 3 types of invocation (see below)
```
Usage:
 
    elf2tab --stack[=<stacksize>] --app-heap[=<appheap>] --kernel-heap[=<kernelheap>] <Elffile>...
    elf2tab --minimum-ram-size[=<size>] <Elffile>...
    elf2tab <Elffile>...
    
...
```
The third one is effectively the same as the first one if the user chooses 0 for all three of the options as value, right?

Nicoretti:
Hi @bradjc and @alevy ,

I just updated the CLI, working in your comments and adjusting some things according to commonly seen CLI styles like  git e.g. arguments `<arg>` instead of `ARG` ...

screen shot 2018-12-03 at 21 01 19

if you have any additions just let me know.

greez
Nico

alevy:
@Nicoretti This seems almost exactly what I'd want, except I think it should be "--minimum-ram-size" instead of "--minimum-stack-size".

@bradjc thoughts?

bradjc:
Right, `--minimum-ram-size` makes sense.

What are the default set to?

```
STACK_SIZE       ?= 2048
APP_HEAP_SIZE    ?= 1024
KERNEL_HEAP_SIZE ?= 1024
```
?

alevy:
@bradjc They are set to that in `libtock-c` Makefile's which passes those arguments explicitly to elf2tab.

I think that's good. elf2tab should default to 0 (which means it'll reserve enough RAM for each of the in-memory sections in the ELF)

bradjc:
I think it should pick reasonable defaults, as the user never wants an app that would be impossible to execute. Also, at that point the options are no longer optional in any real scenario.

Nicoretti:
> Specifically to your points, I think they boil down to "it's hard to guess at build time how much memory will be needed for the stack/app heap/grant region." That's true, but it's not necessarily easier when calling elf2tab, which is also done when "compiling" a process.

@alevy 
I agree, and I think for bigger or more complex applications, there is no way, the application developer can skip the task of putting some reasoning in the memory usage issue. 

@alevy , @bradjc I have built and run the `libtock-rs` **blink example** with my current `elf2tab` version (where all settings default to 0) and put it onto the hail board -> **it is working**

screen shot 2018-12-04 at 21 01 58
screen shot 2018-12-04 at 21 01 22

img_2754


Maybe the take on what the defaults should accomplish, should be something like this:

* The defaults should enable small applications (small apps, examples etc.) to just work out of the box

**Reasoning:** At some point an application developer needs to think about the memory usage. But you don't want to put this burden onto new users and beginners (so they won't be overwhelmed).
They already have to deal with a new os, api, etc. so give them a chance to learn and do one step at the time (as their apps grow they should and need to think about memory usage).

@bradjc is there a reason why we shouldn't put the all `0` defaults to the test?
In case they are not ideal, we can adjust them based on concrete feedback/issues with this defaults.

bradjc:
I'm not too concerned about exactly how `elf2tab` works, or how the division of effort exists with the rest of the build system, but I do feel strongly that elf2tab shouldn't naively generate binaries that have no chance of working. For example, if the ld file doesn't have a stack section, then, unless the user specifies no stack to elf2tab, elf2tab should just do the right thing and at least allocate some memory for the stack.

Nicoretti:
@bradjc this is a legit point, my naive assumption have been that default linker script(s) probably have a .stack section defined and for those who mess with the linker script (provide their own), I should be knowledgeable enough to know that this might have implications along the way (elf2tab, tockloader, ...). But I definitely you have a legit point, so whats the best take on it in our opinion?
* Taking a `default != 0` for the stack, if so which one?
* Taking a `default !=0` for all the parameters, if so which one(s)?

@alevy whats your take on this?

Nicoretti:
Hi @bradjc and @alevy,

just wanted to know if there is still interest in getting this PR in? If so let me know so i'll by resolving the  conflict, if not feel free to just close this PR.

greez
Nico

alevy:
@Nicoretti I'm still in favor of this PR, and I don't think the disagreement between @bradjc, you, and me is important enough to block it. I'm fine going with @bradjc's version and if it matters enough for me at some point I can propose a separate PR to change that behavior.

Nicoretti:
@bradjc, @alevy fixed conflicts and restored old defaults 

```
STACK_SIZE             = 2048
APP_HEAP_SIZE      = 1024
KERNEL_HEAP_SIZE  = 1024
```

alevy:
@bradjc this look OK to you now? I'm happy with it.

PR #9, opened 2019-04-14
Driver Permissions Policy companion

ssands1:
### Pull Request Overview

This PR goes hand in hand with [tock PR #1300](https://github.com/tock/tock/pull/1300). This update to elf2tab allows elf2tab to write permission bytes to TAB files. Tock OS will check these bytes to see if an application is allowed permission to any given driver. Permission bytes form a bitmap, with a driver allowed if its corresponding bit is set to 1. Permission bits can be flipped using tockloader in a coming PR.

Merging this will cause a warning to appear when using the current version of tockloader: 
```
Warning: Unknown TLV block in TBF header: 5.
Warning: You might want to update tockloader.
```
However, this PR shouldn't cause any errors; the permissions header will just be ignored if the version of tockloader doesn't support the feature.

### Help wanted

To know how many bytes to leave blank for permissions, elf2tab needs to know the number of drivers. This results in a `const NUM_DRIVERS` at the top of header.rs which must be incremented each time a driver is add in tock. Is there a better way?

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -61,7 +61,7 @@ struct TbfHeaderWriteableFlashRegion {
 #[derive(Clone, Copy, Debug)]
 struct TbfHeaderPerm {
     base: TbfHeaderTlv,
-    permissions: [u32; DEFAULT_PERMS_LENGTH],
+    permissions: &'static [u32],
 }
 
 impl fmt::Display for TbfHeaderBase {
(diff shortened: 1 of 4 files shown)


src/header.rs
```diff
@@ -8,6 +8,14 @@ use std::mem;
 use std::vec;
 use util;
 
+// number of drivers in tock/capsules/src/driver.rs
+const NUM_DRIVERS: usize = 30;
+// default: no permissions
+const DEFAULT_PERMS: u32 = 0b101;
```
alevy: why does `0b101` mean no permissions?

ssands1: Whoops, left over from testing...

Issue #10, opened 2019-09-06
Multiple entry points detected on Cortex M0 ELF


rajivr:
I've run into an issue where `elf2tab` is detecting multiple entry points in a Cortex M0 binary, resulting in an invalid `init_fn_offset` value.

```
elf2tab -p "app" -v -o "target/tab/"app".tab" "target/tab/"app"/cortex-m0.elf" --stack 2048 --app-heap 4096 --kernel-heap 1024 --protected-region-size=64
Min RAM size from sections in ELF: 536 bytes
Number of writeable flash regions: 0
Including the .crt0_header section at offset 64 (0x40)
Entry point is in .text section
Including the .text section at offset 128 (0x80)
Including the .stack section at offset 20692 (0x50d4)
Including the .data section at offset 22740 (0x58d4)
Entry point is in .debug_str section
Entry point is in .debug_info section
Entry point is in .debug_line section
Searching for .rel.X sections to add.
TBF Header:
               version:        2        0x2
           header_size:       40       0x28
            total_size:    32768     0x8000
                 flags:        1        0x1

        init_fn_offset:   154437    0x25B45
        protected_size:       24       0x18
      minimum_ram_size:     7704     0x1E18
```

While I've not had time to investigate the root cause of the issue (as I'll need to relearn a fair bit about ELF), I am using a somewhat simple (and possibly broken workaround).

```
diff --git a/src/main.rs b/src/main.rs
index e42474c..d2c7c16 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -214,6 +214,9 @@ fn elf_to_tbf(
     // Need a place to put the app sections before we know the true TBF header.
     let mut binary: Vec<u8> = vec![0; protected_region_size as usize - header_length];
 
+    // HACK
+    let mut entry_point_found = false;
+
     // Iterate the sections in the ELF file and add them to the binary as needed
     for s in &sections_sort {
         let section = &input.sections[s.0];
@@ -222,7 +225,10 @@ fn elf_to_tbf(
         // is, then we need to calculate the correct init_fn_offset.
         if input.ehdr.entry >= section.shdr.addr
             && input.ehdr.entry < (section.shdr.addr + section.shdr.size)
+            && !entry_point_found
         {
+            entry_point_found = true;
+
             if verbose {
                 println!("Entry point is in {} section", section.shdr.name);
             }
```

ppannuto:
> Entry point is in .text section
> Entry point is in .debug_str section
[...]

PR #11, opened 2019-09-07
src/main.rs: Ignore spurious entry points in `debug` section

rajivr:
As suggested [here](https://github.com/tock/elf2tab/issues/10#issuecomment-529001887).

Please review.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -214,6 +214,8 @@ fn elf_to_tbf(
     // Need a place to put the app sections before we know the true TBF header.
     let mut binary: Vec<u8> = vec![0; protected_region_size as usize - header_length];
 
+    let mut entry_point_found = false;
+
     // Iterate the sections in the ELF file and add them to the binary as needed
     for s in &sections_sort {
         let section = &input.sections[s.0];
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -222,7 +224,14 @@ fn elf_to_tbf(
         // is, then we need to calculate the correct init_fn_offset.
         if input.ehdr.entry >= section.shdr.addr
             && input.ehdr.entry < (section.shdr.addr + section.shdr.size)
+            && (section.shdr.name.find("debug")).is_none()
         {
+            // panic in case we detect entry point in multiple sections.
+            if entry_point_found {
```
ppannuto: Might be nicer to make this a reference to / copy of the first place an entry point was found, that way the error message can print both places entry points were found for easier debugging.

Assuming that's easy, it'd be a nice change; failing that I'm happy to merge this as-is 👍

rajivr: In the verbose mode, we currently print the first entry point, followed by a panic (in the subsequent iteration of the loop). Wouldn't that accomplish the same thing?

```
            // panic in case we detect entry point in multiple sections.
            if entry_point_found {
                panic!("Duplicate entry point in {} section", section.shdr.name);
            }
            entry_point_found = true;

            if verbose {
                println!("Entry point is in {} section", section.shdr.name);
            }
```

PR #14, opened 2020-03-31
Update package manifest with correct repo url

JayKickliter:
Clicking the `Repository` link at https://crates.io/crates/elf2tab should bring you to this repo, and not the main Tock repo.


brghena:
Thanks @JayKickliter! This is good to merge, but we would either need to update to 4.0.1 to push the URL change or else it can sit waiting in this repo until the next update to elf2tab, which could be a long time off.

JayKickliter:
I'm not sure where `4.0.1` comes from. The manifest current shows: ```version = "0.5.0-dev"```. Either way, this can wait until the next release.

brghena:
Oops, I misread crates.io. I meant to say we could potentially release this as 0.4.1. But since already have a bunch of commits into 0.5.0-dev, that's not a particularly good option.

We'll merge this now and it'll update crates.io whenever a 0.5.0 release happens instead.

PR #16, opened 2020-04-28
Implement a --deterministic mode for reproducibility.

### LINKED ISSUE #15: Reproducible elf2tab
TAB files created by `elf2tab` are currently not reproducible (https://github.com/google/OpenSK/pull/94).

Reasons for it are:
- The build date stored in metadata. This could be fixed by adding some `--deterministic` flag to `elf2tab`. https://github.com/tock/elf2tab/blob/f2220b6a72e01789e2d935cfc3696fb45d7e03c5/src/main.rs#L25-L33
- The tar file contains filesystem metadata about the files read from files. This could be solved with `tar::Builder::mode(Deterministic)` (https://docs.rs/tar/0.4.26/tar/struct.Builder.html#method.mode) https://github.com/tock/elf2tab/blob/f2220b6a72e01789e2d935cfc3696fb45d7e03c5/src/main.rs#L37

gendx:
This pull request fixes #15.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,13 +25,15 @@ FLAGS:
     -v, --verbose    Be verbose
 
 OPTIONS:
-    -o, --output-file <filename>             output file name [default: TockApp.tab]
-        --app-heap <heap-size>               in bytes [default: 1024]
-        --kernel-heap <kernel-heap-size>     in bytes [default: 1024]
-        --minimum-ram-size <min-ram-size>    in bytes
-    -p, --package-name <pkg-name>            package name
-        --protected-region-size <protected-region-size>    Size of the protected region (including headers)
-        --stack <stack-size>                 in bytes [default: 2048]
+        --deterministic                      Produce a deterministic TAB file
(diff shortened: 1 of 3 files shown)


gendx:
As mentioned in https://docs.rs/tar/0.4.26/tar/enum.HeaderMode.html#variant.Deterministic

> Only metadata that is directly relevant to the identity of a file will be included. In particular, ownership and mod/access times are excluded.

ppannuto:
I don't see any harm in unconditionally turning on Deterministic for the tar file; will let this sit for others to weigh in on the date question, but my current disposition is for elf2tab to just always be deterministic.

bradjc:
I think it might be useful to leave the TAB metadata file open to including build-specific data in the future. Having a `--deterministic` flag gives us that flexibility.

I do think that the `--deterministic` option can just omit the build-date entry, however.

gendx:
> I do think that the `--deterministic` option can just omit the build-date entry, however.

Done.

gendx:
Is there anything blocking this? A new release on crates.io would also be helpful after this pull request.

PR #17, opened 2020-04-28
Apply various cleanups: formatting, build warnings, Clippy

gendx:
This pull request applies various cleanups:
- `cargo fmt`.
- Fix a few build warnings about missing `dyn` references (refactored to use static dispatch) and unneeded parentheses.
- Various `cargo clippy` warnings (including `write_all` instead of `write` for correctness). Printing is simplified to use `println` and avoid duplicating format arguments.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -203,7 +203,10 @@ impl TbfHeader {
         }
 
         // Return the length by generating the header and seeing how long it is.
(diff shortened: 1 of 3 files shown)


gendx:
Can we merge this so that I can update #16 to resolve the merge conflicts between the two pull requests?

PR #18, opened 2020-05-07
revert -n to -p change

bradjc:
This re-establishes `-n` flag so that libtock-c builds. I don't think such a small change is worth making an API breaking change for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ FLAGS:
 OPTIONS:
         --deterministic                      Produce a deterministic TAB file
     -o, --output-file <filename>             Output file name [default: TockApp.tab]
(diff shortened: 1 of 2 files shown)


ppannuto:
I had to look up in blame when this changed -- it's been this way since December 2018.

Given that it hasn't caused any issues for anyone until now, I might be more inclined to update libtock-c. I do agree with the rationale from the dec'18 change:
> * Change short option name of package-name form 'n' to 'p' to comply with commonly used convention to use the first letter of the long option for the short option

bradjc:
Well that is when the commit was authored, it wasn't merged until later. It hasn't been in a release, and I don't think it has caused an issue because no one is running elf2tab v0.5.0-dev, otherwise libtock-c wouldn't compile.

PR #21, opened 2020-05-24
Update to 2018 edition

torfmaster:
# Summary

This PR updates elf2tab to the 2018 edition. Only the imports have been amended to match new "path clarity" changes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,24 +1,18 @@
-extern crate chrono;
-extern crate elf;
-extern crate tar;
-#[macro_use]
-extern crate structopt;
-
+use cmdline::Opt;
 use std::cmp;
 use std::fmt::Write as fmtwrite;
 use std::fs;
 use std::io;
 use std::io::{Seek, Write};
 use std::mem;
-
-#[macro_use]
-mod util;
-mod cmdline;
-mod header;
 use structopt::StructOpt;
 
+mod cmdline;
+mod header;
+mod util;
+
 fn main() {
-    let opt = cmdline::Opt::from_args();
+    let opt = Opt::from_args();
 
     let package_name = opt
         .package_name
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,3 @@
-use cmdline::Opt;
 use std::cmp;
 use std::fmt::Write as fmtwrite;
 use std::fs;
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -1,24 +1,19 @@
...
-#[macro_use]
-mod util;
 mod cmdline;
 mod header;
-use structopt::StructOpt;
+mod util;
 
 fn main() {
-    let opt = cmdline::Opt::from_args();
+    let opt = Opt::from_args();
```
bradjc, line 16: Undo

src/main.rs
```diff
@@ -1,24 +1,19 @@
-extern crate chrono;
-extern crate elf;
-extern crate tar;
-#[macro_use]
-extern crate structopt;
-
+use cmdline::Opt;
```
bradjc, line 1: Using the full path is nice as it tells people right in the source where the very short and not unique "Opt" object is from.

torfmaster:
> Was this intended to be only the first commit? I don't understand the alignment changes. What is wrong with macros?

Sorry, I forgot to comment on that. I tried to add some tests to the crate however, I found out that there is some complexity hidden in the Macros due to hidden conversions (usize to u32 etc.). I wanted to make this explicit using generics but ended up using Numeric traits. Finally, I decided to do this change making the conversions explicit. If found it uncontroversial so I included it. I can split out this change and make it a separate "add tests" PR if you like.

src/main.rs
bradjc, line 16:
```suggestion
    let opt = cmdline::Opt::from_args();
```

src/main.rs
bradjc, line 1:
```suggestion
```

torfmaster: Fixed.

src/main.rs
torfmaster: Fixed.

torfmaster:
> Gotcha, that seems reasonable. Is the cmdline change necessary? Otherwise good to go.

Sorry, I didn't get the answer. Shall I split the PR into two?

torfmaster:
I would suggest to close #3 after the tests are merged, although the test coverage could be better... (I hope I set a trend here :)

PR #22, opened 2020-05-28
Add support for Fixed Addresses TLV

bradjc:
This adds support for the new TLV that allows for specifying where in ram and flash the app needs be. The addresses are inferred from the ELF. If the addresses are flash=0x80000000 and ram=0x00000000 then we assume the app supports PIC and we omit the fixed address TLV. If the addresses are different, then we add the header.

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -1,5 +1,6 @@
 //! Command line parser setup for elf2tab.
 
+use std::num::ParseIntError;
 use std::path::PathBuf;
 
 use structopt;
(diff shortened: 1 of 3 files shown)


bradjc:
@tock/core-wg

bradjc:
Arg ok this doesn't work for IPC apps (probably apps that do something special with global buffers).

bradjc:
Updated, and this now works for the IPC libtock-c apps.

ppannuto:
Could we make the PIC-guessing easier by being more explicit at the linking step?

e.g. currently the only things that support PIC are libtock-c apps, all of which necessarily have a section named `.crt0-header` (which happens to be placed at our sentinel value 0x8000_0000). Perhaps we rename that to `.tock-PIC-header`. Or we add a dedicated section called `.tock-PIC-true` to the ELF. Basically have the linker explicitly opt-in stating the app supports PIC?

bradjc:
Well I'm not sure that is easier, that requires changing multiple repositories. But we would still have the code to guess the addresses, is it worth having two separate mechanisms for the same general task?

ppannuto:
> But we would still have the code to guess the addresses

Why? Wouldn't the whole point be that we don't need address guessing code if there is a properly named elf section (if no elf section, no PIC)? In fact, we could even eliminate the "magic" 0x8000_0000 number by making a section in the ELF named `.tock-PIC-address` with one word in it that is the address.

bradjc:
We need to know when PIC _isn't_ supported. I kind of have come to like this approach, since it "just works" without having to update libtock-c (riscv only) and libtock-rs (which I can't build).

bradjc:
@tock/core-wg 

I think this is really important for helping debug RISC-V issues.

PR #23, opened 2020-06-24
Improve how elf2tab parses ELF files

bradjc:
This fixes two things:

1. Calculating how much RAM the app needs based on the ELF sections. Before we stopped after finding any segment that is destined for RAM. This doesn't work for libtock-c now that .stack is its own section. So not only do we consider multiple segments if necessary, but the checking for if it should count towards the amount of RAM required is more robust.
2. Finding fixed addresses. The fixed address in flash is mostly the same, made a little more robust for libtock-rs. RAM however won't work the same way since libtock-rs doesn't generate any segments at the correct address in RAM. Therefore, we use a special symbol the linker file has to specify.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ elf2tab will also automatically add a TBF "fixed addresses" TLV header if it
 finds that the .elf file was compiled for a fixed address in RAM or flash
 instead of being position independent. To detect a fixed flash address, elf2tab
 looks to see if the flash segment is at the dummy flash address for PIC apps or
-not. To detect a fixed RAM address, elf2tab looks for a `_SRAM_ORIGIN` symbol,
+not. To detect a fixed RAM address, elf2tab looks for a `_sram_origin` symbol,
 and checks if the address matches the dummy RAM address for PIC apps or not.
 
 
@@ -127,5 +127,4 @@ Inspecting TABs
 Tockloader can show some details of a .tab file. Simply:
 
     $ tockloader inspect-tab <tab file name>
-
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,14 @@ writeable flash regions, the name of the section should include the string
 their relative address offset included in the TBF header via the
 `TbfHeaderWriteableFlashRegions` TLV.
 
+elf2tab will also automatically add a TBF "fixed addresses" TLV header if it
+finds that the .elf file was compiled for a fixed address in RAM or flash
+instead of being position independent. To detect a fixed flash address, elf2tab
+looks to see if the flash segment is at the dummy flash address for PIC apps or
+not. To detect a fixed RAM address, elf2tab looks for a `_SRAM_ORIGIN` symbol,
+and checks if the address matches the dummy RAM address for PIC apps or not.
+
+
 ### Creating the TAB file
 
 After generating the program binary and TBF header for each .elf file specified
--- a/README.md
+++ b/README.md
@@ -108,7 +108,8 @@ finds that the .elf file was compiled for a fixed address in RAM or flash
 instead of being position independent. To detect a fixed flash address, elf2tab
 looks to see if the flash segment is at the dummy flash address for PIC apps or
 not. To detect a fixed RAM address, elf2tab looks for a `_sram_origin` symbol,
-and checks if the address matches the dummy RAM address for PIC apps or not.
+and if it exists checks if the address matches the dummy RAM address for PIC
+apps or not.
 
 
 ### Creating the TAB file
@@ -127,4 +128,3 @@ Inspecting TABs
 Tockloader can show some details of a .tab file. Simply:
 
     $ tockloader inspect-tab <tab file name>
-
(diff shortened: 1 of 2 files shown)


alistair23:
Why are the RAM segments different for `libtock-rs`?

bradjc:
.stack is set to noload and does not generate a segment.

README.md
```diff
@@ -103,6 +103,14 @@ writeable flash regions, the name of the section should include the string
 their relative address offset included in the TBF header via the
 `TbfHeaderWriteableFlashRegions` TLV.
 
+elf2tab will also automatically add a TBF "fixed addresses" TLV header if it
+finds that the .elf file was compiled for a fixed address in RAM or flash
+instead of being position independent. To detect a fixed flash address, elf2tab
+looks to see if the flash segment is at the dummy flash address for PIC apps or
+not. To detect a fixed RAM address, elf2tab looks for a `_SRAM_ORIGIN` symbol,
```
bradjc, line 110:
```suggestion
not. To detect a fixed RAM address, elf2tab looks for a `_sram_origin` symbol,
```

src/main.rs
```diff
@@ -151,48 +161,76 @@ fn elf_to_tbf<W: Write>(
...
+    // For RAM, we can't quite use the ELF file in the same way as with flash.
+    // Since nothing _actually_ has to be loaded into RAM, the ELF file does not
+    // have to keep track of the first address in RAM. Further, Tock apps
+    // typically put a .stack section at the beginning of RAM, and the stack is
+    // just a memory holder and doesn't contain any actual data. The start of
+    // RAM address will almost certainly exist somewhere in the ELF, but
+    // reliably extracting it from different ELFs linked with different
+    // toolchains from different linker scripts would I think require resorting
+    // to a bit of heuristics and guessing. To avoid the potential issues there,
+    // we instead require that a `_SRAM_ORIGIN` symbol be present to explicitly
```
bradjc, line 179:
```suggestion
    // we instead require that a `_sram_origin` symbol be present to explicitly
```

src/main.rs
```diff
@@ -213,9 +251,27 @@ fn elf_to_tbf<W: Write>(
...
-        fixed_address_ram = None;
-    }
+
+    // Do RAM address.
+    // Get all symbols in the symbol table section if it exists.
+    let section_symtab = input.sections.iter().find(|s| s.shdr.name == ".symtab");
+    section_symtab.map(|s_symtab| {
+        let symbols = input.get_symbols(s_symtab);
+        symbols.ok().map(|syms| {
+            // We are looking for the `_SRAM_ORIGIN` symbol and its value.
```
bradjc, line 261:
```suggestion
            // We are looking for the `_sram_origin` symbol and its value.
```

src/main.rs
```diff
@@ -213,9 +251,27 @@ fn elf_to_tbf<W: Write>(
...
+            // If it exists, we try to use it. Otherwise, we just do not try
+            // to find a fixed RAM address.
+            let sram_origin_symbol = syms.iter().find(|sy| sy.name == "_SRAM_ORIGIN");
```
bradjc, line 264:
```suggestion
            let sram_origin_symbol = syms.iter().find(|sy| sy.name == "_sram_origin");
```

README.md
```diff
@@ -103,6 +103,14 @@ writeable flash regions, the name of the section should include the string
 their relative address offset included in the TBF header via the
 `TbfHeaderWriteableFlashRegions` TLV.
 
+elf2tab will also automatically add a TBF "fixed addresses" TLV header if it
+finds that the .elf file was compiled for a fixed address in RAM or flash
+instead of being position independent. To detect a fixed flash address, elf2tab
+looks to see if the flash segment is at the dummy flash address for PIC apps or
+not. To detect a fixed RAM address, elf2tab looks for a `_sram_origin` symbol,
+and checks if the address matches the dummy RAM address for PIC apps or not.
```
ppannuto, line 111: What does elf2tab do if there is neither a `_sram_origin` symbol nor does the PIC address match? (I'll read the code in a second, but I think the intent should probably be specified here (i.e. will/should it error?))

src/main.rs
```diff
@@ -151,48 +161,76 @@ fn elf_to_tbf<W: Write>(
...
+    //
+    // In both cases we check to see if the address matches our expected PIC
+    // addresses:
+    // - RAM: 0x00000000
+    // - flash: 0x80000000
+    //
+    // These addresses are a Tock convention and enables PIC fixups to be done
+    // by the app when it first starts. If for some reason an app is PIC and
+    // wants to use different dummy PIC addresses, then this logic will have to
+    // be updated.
```
ppannuto, line 190: Not for this PR, but I think this logic shouldn't change; rather these constant should just be exposed as flags.

README.md
bradjc: No, that would break backwards compatibility. elf2tab just does what it does now: nothing.

bradjc, line 111:
```suggestion
and if it exists checks if the address matches the dummy RAM address for PIC apps or not.
```

ppannuto:
Haha, there's no bors ... or even tests (😢) ... in this repo is there

bradjc:
There are tests! But bors doesn't do anything if there is only one PR/week or month or year.

PR #24, opened 2020-07-06
Estimate a protected region size for fixed address apps

bradjc:
Part of the difficulty of using fixed address apps is that the start of the TBF also has to fall at a certain address so the kernel can find the app. We worked around this issue with the `--protected-region-size` argument to elf2tab so that the tbf start can be set to a fixed offset before the application binary. The downside is this applies to all apps being inserted into the TAB, even if they are PIC.

This PR lets elf2tab estimate a reasonable length for the offset of the application binary in the TBF file, if the TBF is compiled for a fixed address in flash. This makes the raw TBF binary a little easier to use, as in most cases it can be directly flashed to the board. elf2tab has no way of actually knowing where the TBF will be placed, and how large the protected region should be, so this is only a guess. Tockloader will be able to handle the other cases by dynamically expanding the protected region size once it knows all of the relevant addresses when flashing the application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,19 @@ not. To detect a fixed RAM address, elf2tab looks for a `_sram_origin` symbol,
 and if it exists checks if the address matches the dummy RAM address for PIC
 apps or not.
 
+elf2tab has to choose a length for the protected region after the TBF header and
+before the start of the actual application binary. Normally, this defaults to 0.
+It can be fixed for all TBFs in the TAB using the command line argument
+`--protected-region-size` (which takes as an argument entire size before the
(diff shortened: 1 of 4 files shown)


bradjc:
My goal is to maximize the chances things just work, and I think that means putting logic in every layer to cover different platforms and use cases. I had thought tockloader would provide enough value that it would be the default tool, but I understand that its also an unfamiliar tool to anyone new to Tock. I can also see the appeal of the same binary on your computer being what is programmed on to the board.

bradjc:
I've been using this patch for a while, and verified that it works with arty-e21 and multiple riscv apps.

PR #25, opened 2020-09-15
Added X86 relocation

alexandruradovici:
This PR adds X86 relocation for https://github.com/tock/tock/pull/2111.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "elf2tab"
-version = "0.6.0-dev"
+version = "0.7.0-dev"
 dependencies = [
  "chrono 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


bradjc:
I tried adding a commit to this PR, but check out: https://github.com/tock/elf2tab/commit/909eb8deae23ad0c4b644b166e985521e33c789d

I think RISC-V uses rela locations as well, so it makes sense to just have them be on by default in elf2tab. Let me know what you think.

alexandruradovici:
> I tried adding a commit to this PR, but check out: [909eb8d](https://github.com/tock/elf2tab/commit/909eb8deae23ad0c4b644b166e985521e33c789d)
> 
Looks good.
> I think RISC-V uses rela locations as well, so it makes sense to just have them be on by default in elf2tab. Let me know what you think.
I don't see why not, I will try a few more names (compare more strings), but I don't see a real issue.

bradjc:
Please re-open if there is still work on this.

PR #26, opened 2020-10-07
Update Cargo.lock for v0.6.0 release

lschuermann:
Unfortunately the v0.6.0 release does not contain an up-to-date `Cargo.lock` for package managers such as Nix (which rely on those) or "reproducible" `--frozen` Cargo builds in general.

To be able to use elf2tab as part of the Nix package manager I'd kindly ask to create a patch release `v0.6.1` with a valid `Cargo.lock`.

This should probably not be merged into master. Since the GitHub UI won't allow me to propose merging to a new branch, I'd suggest to [apply the patch of this PR](https://patch-diff.githubusercontent.com/raw/tock/elf2tab/pull/26.patch) locally on a new temporary branch starting at the `v0.6.0` release, then creating a tag on the applied commit and pushing that tag.

Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,345 +4,338 @@
 name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
 dependencies = [
- "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi",
 ]
 
 [[package]]
 name = "atty"
-version = "0.2.10"
+version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
- "libc 0.2.40 (registry+https://github.com/rust-lang/crates.io-index)",
- "termion 1.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hermit-abi",
+ "libc",
+ "winapi",
 ]
 
 [[package]]
-name = "bitflags"
-version = "1.0.3"
+name = "autocfg"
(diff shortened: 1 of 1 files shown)


hudson-ayers:
Why do you not recommend merging this into master? I am not particularly familiar with how `Cargo.lock` works

lschuermann:
`elf2tab` might have changed its dependencies or versions, so to be sure that this patch applies (not only in the Git but also in the Rust sense) I recommend to just branch off of the v0.6.0 release and create a patch release containing this change exclusively. An alternative would be to discard this and release v0.7.0 with an according Cargo.lock.

Nix (and other package managers) just need a matching Cargo.lock to build a package out of it, so updating it prior to the release is required.

lschuermann:
@bradjc ping :)

hudson-ayers:
I pushed a v0.6.1 tag with an updated `Cargo.lock`. It would probably be nice to move to something more like the main Tock repo, where `Cargo.lock` is not checked into the repo at all, and just generated by the build system.

Let me know if for any reason this does not work.

lschuermann:
`Cargo.lock` _should_ be included in a release as it pins the dependencies to versions with fixed hashes, so that the build becomes mostly reproducible. Also, `crates.io` allows you to download a version of a crate even if it has been revoked, provided you have the hash (stored in Cargo.lock). Anyone packaging `elf2tab` for their OS will want to have a Cargo.lock in the release.

It's sufficient to not check in a new `Cargo.lock` for every commit, but prior to each release it should be updated accordingly.

You can read more in [the cargo guide](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html), the most important statement being "_If you’re building a non-end product, such as a rust library that other rust packages will depend on, put Cargo.lock in your .gitignore. If you’re building an end product, which are executable like command-line tool or an application, or a system library with crate-type of staticlib or cdylib, check Cargo.lock into git._"

Whether Tock should count as an end product is questionable. It's not a pressing issue for me. If someone uses Tock in production they'll likely have their own board-crate, pulling in the kernel as a cargo dependency, so they can have their own `Cargo.lock`.

hudson-ayers:
Thanks for the explanation, I didn't know that

lschuermann:
The `elf2tab` Nix build now works, thanks!

PR #27, opened 2020-10-20
Add support for an app_id

alistair23:
Add support for an app_id. This can be speicified via the command line
with the --app-id argument. If one isn't specified one will be generated
based on a hash of the name.

App IDs will eventually be used by the kernel to track apps across
updates, layout changes and name changes. They should be unique and not
change during the lifetime of an app.


src/header.rs
```diff
@@ -206,6 +214,18 @@ impl TbfHeader {
...
+
+        // Generate an AppID from the package name
+        if app_id.is_none() {
+            let hash = hasher.finalize();
+            self.hdr_main.app_id = hash[0] as u32
+                | (hash[1] as u32) << 8
+                | (hash[2] as u32) << 16
+                | (hash[3] as u32) << 24;
+        } else {
+            self.hdr_main.app_id = app_id.unwrap();
```
hudson-ayers, line 228: If the package name does not exist this will use the default value inside the hasher, which could lead to two unnamed packages having the same AppId. The same problem would exist for two packages with the same name being flashed. This seems like it could be a pretty big problem if the kernel assumed that passed appids do not collide.

alistair23: Hmm... Good point.

The goal was so that casual users would generally end up with app-id's that are constant over updates and layout changes.

I could swap this to just be a random number, but then apps will never have the same ID unless the user manually specifies it. I'm not sure if that matters much though.

I'll look into this more

lschuermann:
tl;dr: I like this approach, we should refrain from using sections of the app in the hash which might change to allow for application id persistency.

I think that [this discussion on the mailing list](https://groups.google.com/forum/#!topic/tock-dev/aduN7fHWXdI) and [this email](https://groups.google.com/d/msg/tock-dev/aduN7fHWXdI/2jYwTSMzBgAJ) relevant. In my opinion we should differentiate between
- application ids
- verifiable ids

where an _application id_ is a small (native data-width) size id used throughout the kernel during runtime for efficiency reasons and a _verifiable id_ is an identifier based on cryptographic processes.

I like that this introduces the notion of an _application id_ or more specifically, a default format for those (after all, the _application ids_ **do not have to** be based on the TBF header and **can** be set arbitrarily using a nonstandard method to load the apps, thereby can be based on some kind of _verifiable id_ if required) .

We should in my opinion make sure that the _application ids_ are dependent on a constant, user specified value (such as the package name, as you proposed). Using a hash of other parts of the application which are not user defined, but depend on the application's elf sections would mix the concepts of _verifiable ids_ and _application ids_, and make any persistent resource allocation during updates etc. hard if not impossible.

As a conclusion I think that application ids do not necessarily have to be unique over processes on a kernel, as for that we have the (now quite confusingly named) `AppId` already:
https://github.com/tock/tock/blob/07bd824508b420c0daa1133e2a3bfa2550718306/kernel/src/process.rs#L776-L780

Rather, the _application id_ would be akin to the concept of namespaces under Linux, which controls resource allocation (for instance file ownership, in the context of a filesystem). Using this definition an empty package name might be a non-issue, since those processes would be running in the "default namespace".

src/header.rs
lschuermann: > The goal was so that casual users would generally end up with app-id's that are constant over updates and layout changes.

I think that this is a large part of why we want to have application ids in the first place. I'd refrain from using something that cannot be user controlled. See [my comment](https://github.com/tock/elf2tab/pull/27#issuecomment-715112111) for a further explanation of my reasoning.

bradjc (changes requested):
Block because at minimum we can't change the TBF main field, we have to add a TLV.

I would like to see us move quickly on a unique, persistent ID for applications. However, I do not think that elf2tab should insert this because that has the implication that it will be a global identifier. Since elf2tab cannot enforce that the identifier will be globally unique, it can lead to confusion if a collision happens. I would rather the tool that loads applications be responsible for setting/checking the identifier.

alistair23:
I'm closing this as it seems no one is really on board with an AppiID in elf2tab.

PR #29, opened 2021-03-01
Add flags to TAB metadata.toml file

bradjc:
This PR updates the metadata.toml file that elf2tab generates. Specifically:

- It adds a command line flag `--supported-boards` so that apps can specify which boards they are compatible with (if there happens to be a particular restriction). For example `--supported-boards hail,imix` would then add that information to the metadata file in the TAB, and then tockloader can check and give an error to the user if they try to install that app on a board not named hail or imix.

- ~~It adds a second command line flag `--tock-kernel-version`. Again this adds an entry to the metadata file, so that an app can specify which kernel ABI version it requires and tockloader can help check. My intention is that a 2.0-compliant userspace would add `elf2tab ... --tock-kernel-version 2` to its build system.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ OPTIONS:
         --read_ids <read_ids>...                           Storage IDs that this app is allowed to read
         --stack <stack-size>                               in bytes [default: 2048]
         --write_id <write_id>                              A storage ID used for writing data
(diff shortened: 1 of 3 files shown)


hudson-ayers (approved):
Not sure if you still wanted this but the implementation looks fine to me

bradjc:
Can we merge this now?

PR #30, opened 2021-07-18
Add kernel version header implementation

alexandruradovici:
This PR adds the `KernelVersion` header implementation defined in https://github.com/tock/tock/pull/2669.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "elf2tab"
-version = "0.7.0-dev"
+version = "0.8.0-dev"
```
bradjc, line 3:
```suggestion
version = "0.7.0-dev"
```

alexandruradovici: Shouldn't we increase the version due to this new header?

bradjc: 0.7 is already an increase.

alexandruradovici: Got it now :)

hudson-ayers (approved):
I don't see anything in the updated diff that looks like it could interfere with Alistair's work that was merged before this, I think it is good to go

bradjc (approved):

PR #33, opened 2021-08-09
Make binary padding to power of two optional, only use for cortex-m

bradjc:
elf2tab was originally written for only cortex-m apps, and since those have always needed to be padded to a power of two, the functionality to add the padding has always been baked in. Nowadays, not all apps need to be padded (RISC-V boards do not have the same requirement), and we don't necessarily want to force long apps in all cases.

This changes elf2tab to only pad cortex-m apps.

We _could_ not pad any apps, and leave the padding up to tockloader. Tockloader can handle this just fine. However, that would make it difficult to manually flash .tbf files to cortex-m boards without tockloader, so it is probably better to have both elf2tab and tockloader support the padding.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -91,6 +91,17 @@ fn main() {
             .open(tbf_path.clone())
             .unwrap();
 
+        // Adding padding to the end of cortex-m apps. Check for a cortex-m app
+        // because the start of the elf filename will start with "cortex".
+        //
+        // RISC-V apps do not need to be sized to power of two.
+        let add_trailing_padding = elf_path
+            .file_name()
+            .unwrap_or(std::ffi::OsStr::new(""))
+            .to_str()
+            .unwrap_or("")
+            .starts_with("cortex");
+
         // Do the conversion to a tock binary.
         if opt.verbose {
             println!("Creating {:?}", tbf_path);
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
This looks great, though I suspect there is some risk that we will one day add an architecture that which is neither cortex nor riscv, and the logic for determining whether to add padding will automatically treat that architecture as not needing padding. It might be slightly safer for the default to assume padding requirements, if there is an easy way to know that an app is for a risc-v board based on the elf filename.

src/main.rs
```diff
@@ -91,6 +91,17 @@ fn main() {
...
+        // Adding padding to the end of cortex-m apps. Check for a cortex-m app
+        // because the start of the elf filename will start with "cortex".
+        //
+        // RISC-V apps do not need to be sized to power of two.
+        let add_trailing_padding = elf_path
+            .file_name()
+            .unwrap_or(std::ffi::OsStr::new(""))
+            .to_str()
+            .unwrap_or("")
+            .starts_with("cortex");
```
ppannuto, line 103: Doing this based on filename feels pretty precarious, isn't the architecture metadata in the elf file that this is already parsing?

```
$ file cortex-m0/cortex-m0.elf
cortex-m0/cortex-m0.elf: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, with debug_info, not stripped
```

Even the `file` utility knows this is for ARM..

alistair23 (approved):

bradjc:
Ok switched to using the elf header instead of the filename for cortex-m detection.

hudson-ayers (approved):

ppannuto (approved):

PR #37, opened 2021-10-21
Allow creating disabled apps (TBF disabled flag set)

alistair23:
This PR allows creating disabled apps that the Tock kernel can then use to program the [OTBN](https://docs.opentitan.org/hw/ip/otbn/doc/index.html#running-applications-on-otbn)

### DIFF
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -64,6 +64,9 @@ pub struct Opt {
     #[structopt(long = "deterministic", help = "Produce a deterministic TAB file")]
     pub deterministic: bool,
 
+    #[structopt(long = "disable", help = "Mark the app as disabled in the TBF flags")]
+    pub disabled: bool,
+
     #[structopt(
         long = "minimum-ram-size",
         name = "min-ram-size",
--- a/src/cmdline.rs
+++ b/src/cmdline.rs
@@ -92,11 +92,7 @@ pub struct Opt {
     )]
     pub package_name: Option<String>,
 
-    #[structopt(
-        long = "stack",
-        name = "stack-size",
-        help = "in bytes"
-    )]
+    #[structopt(long = "stack", name = "stack-size", help = "in bytes")]
     pub stack_size: Option<u32>,
 
     #[structopt(
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -1,9 +1,9 @@
+use crate::header;
+use crate::util::{self, align_to, amount_alignment_needed};
 use std::cmp;
 use std::io;
 use std::io::Write;
 use std::mem;
-use crate::util::{self, align_to, amount_alignment_needed};
-use crate::header;
 
 /// Convert an ELF file to a TBF (Tock Binary Format) binary file.
 ///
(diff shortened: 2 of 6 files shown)


alistair23:
@bradjc

alistair23:
Ping!

alistair23:
@bradjc  Ping!

src/header.rs
```diff
@@ -357,8 +358,7 @@ impl TbfHeader {
             header_length += mem::size_of::<TbfHeaderKernelVersion>();
         }
 
-        // Flags default to app is enabled.
-        let flags = 0x0000_0001;
+        let flags = if disabled { 0x0000_0000 } else { 0x0000_0001 };
```
cfrantz, line 361: Nit (and not really your error):  The enable/disable bit should be defined as a `const` or `bitflags!` or similar.  A similar definition should be used in [tock](https://github.com/tock/tock/blob/master/libraries/tock-tbf/src/types.rs#L569).

src/main.rs
```diff
@@ -481,9 +481,8 @@ fn elf_to_tbf<W: Write>(
             && input.ehdr.entry < (section.shdr.addr + section.shdr.size)
             && (section.shdr.name.find("debug")).is_none()
         {
-            // panic in case we detect entry point in multiple sections.
             if entry_point_found {
-                panic!("Duplicate entry point in {} section", section.shdr.name);
+                println!("Duplicate entry point in {} section", section.shdr.name);
```
cfrantz, line 485: This behavior should probably be gated on a command-line flag.  For non-otbn apps, this is definately an error.

One could also question whether or not an OTBN app should be packaged directly in a tock app like this or whether an OTBN app should be a data section inside of a "proper" tock app and be given to the OTBN driver by means of a syscall.

I would argue the former is acceptable for hacky development, but the latter should actually be the long-term goal.  An OTBN program _is not_ a tock app and cannot ever execute under tock.  It is, at best, a data payload needed by some other tock app.

cfrantz (commented):
You'll need to rebase this change.  Elf2Tab has had some minor structural (file) changes and has been changed into a bin+lib crate.

src/cmdline.rs
```diff
@@ -35,6 +35,9 @@ pub struct Opt {
     #[structopt(long = "deterministic", help = "Produce a deterministic TAB file")]
     pub deterministic: bool,
 
+    #[structopt(long = "disable", help = "Mark the app as disabled in the TBF flags")]
+    pub disabled: bool,
+
```
bradjc, line 40: We should document this on the readme.

bradjc (commented):
Oh this exposes creating an app as disabled. Great, but we should document that on the readme.

hudson-ayers:
@alistair23 I think this is just waiting on a README update

alistair23:
I updated the REAMD, is there more of an update required?

hudson-ayers (approved):
Sorry, I somehow missed that in the files tab

src/convert.rs
```diff
@@ -368,9 +370,12 @@ pub fn elf_to_tbf<W: Write>(
...
             && (section.shdr.name.find("debug")).is_none()
         {
-            // panic in case we detect entry point in multiple sections.
             if entry_point_found {
-                panic!("Duplicate entry point in {} section", section.shdr.name);
+                if disabled {
+                    println!("Duplicate entry point in {} section", section.shdr.name);
+                } else {
+                    panic!("Duplicate entry point in {} section", section.shdr.name);
+                }
```
bradjc, line 378: I don't understand why having a disabled app changes this logic. A comment here would be helpful.

bradjc (approved):

PR #38, opened 2022-02-14
Use architecture parameter for elf file

alexandruradovici:
This PR makes it possible to tag each elf file with its architecture (`cortex-m...`, `rv32...`).

If the architecture tag is present, the output TBF file's name will be *elf_filename.architecture.tbf* and *architecture.tbf* in the TAB file.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -69,25 +69,7 @@ fn main() {
     // Iterate all input elfs. Convert them to Tock friendly binaries and then
     // add them to the TAB file.
     for elf_file in opt.input {
-        let (tbf_path, architecture) = if let Some(ref architecture) = elf_file.architecture {
-            (
-                elf_file
-                    .path
-                    .with_extension(format!("{}.tbf", architecture)),
-                architecture.clone(),
-            )
-        } else {
-            (
-                elf_file.path.with_extension("tbf"),
-                elf_file
-                    .path
-                    .file_name()
-                    .unwrap()
-                    .to_str()
-                    .unwrap()
-                    .to_string(),
-            )
-        };
+        let tbf_path = elf_file.path.with_extension("tbf");
 
         let elffile = elf::File::open_path(&elf_file.path).expect("Could not open the .elf file.");
 
@@ -142,6 +124,17 @@ fn main() {
 
         // Add the file to the TAB tar file.
         outfile.seek(io::SeekFrom::Start(0)).unwrap();
+        let architecture = if let Some(ref architecture) = elf_file.architecture {
+            architecture.clone()
+        } else {
+            elf_file
+                .path
+                .file_name()
+                .unwrap()
+                .to_str()
+                .unwrap()
+                .to_string()
+        };
         tab.append_file(format!("{}.tbf", architecture), &mut outfile)
             .unwrap();
     }
--- a/src/main.rs
+++ b/src/main.rs
@@ -68,10 +68,26 @@ fn main() {
 
     // Iterate all input elfs. Convert them to Tock friendly binaries and then
     // add them to the TAB file.
-    for elf_path in opt.input {
-        let tbf_path = elf_path.with_extension("tbf");
+    for elf_file in opt.input {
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -68,10 +68,26 @@ fn main() {
 
     // Iterate all input elfs. Convert them to Tock friendly binaries and then
     // add them to the TAB file.
-    for elf_path in opt.input {
-        let tbf_path = elf_path.with_extension("tbf");
+    for elf_file in opt.input {
+        let (tbf_path, architecture) = if let Some(ref architecture) = elf_file.architecture {
```
bradjc, line 72: Why do we need `architecture`? We don't actually need to know the architecture for the rest of the elf2tab conversion.

bradjc: Why doesn't the old logic for 143 work?

alexandruradovici: The  name of the file in the tab needs to be *architecture.tbf*. Libtock-rs uses a *target/tbf* folder where it places all the tbf and tab files, naming them with application_name.{tbf,tab}. This change will make elf2tab name the tbf file with application.architecture.tbf, which is different from what the tab file expects.

As a side note, it will require a change in libtock-c, as the elf files there are named using the architecture. I can add a check here to see if the application name is the same with the architecture (libtock-c's case).

bradjc:
This no longer builds on my machine.

```
   Compiling elf2tab v0.9.0-dev (/Users/bradjc/git/elf2tab)
error: there is no argument named `architecture`
  --> src/main.rs:74:55
   |
74 |                 elf_file.path.with_extension(format!("{architecture}.tbf")),
   |                                                       ^^^^^^^^^^^^^^

error: there is no argument named `architecture`
   --> src/main.rs:143:34
    |
143 |         tab.append_file(format!("{architecture}.tbf"), &mut outfile)
    |                                  ^^^^^^^^^^^^^^

error: aborting due to 2 previous errors

error: could not compile `elf2tab`
```

We need to fix something, if we require a certain version of stable then we need to capture that somehow.

alexandruradovici:
I'm sorry, I used the new format style and I think it needs the newest Rust release. I reverted it to the *parameters* way.

src/main.rs
bradjc: Why is elf2tab changing the name of the elf when creating the tbf? Decoupling the architecture from the filename makes sense to me, but re-introducing new dependencies between elf name and tbf name seems like it has the opposite effect. For example, a build system might reasonably use the name `application-architecture.elf` which would then become `application-architecture.architecture.elf`. I think we should not add the extra name change and just use the same name as what elf2tab is given (with a different extension).

alexandruradovici: I see your point. I'll change it back. 

As a note, this will override applications in libtock-rs if an app is compiled for multiple architectures.

bradjc: > As a note, this will override applications in libtock-rs if an app is compiled for multiple architectures.

Right, if the same elf is used for multiple architectures. That could be a problem (although I think it is separate from this one). I think we should revisit this if we want to support one elf -> multiple architectures, but right now I think either we should a) let the compiler take advantage of a specific architecture and _not_ share elfs across different versions of an arch (preferred), or b) make this a tockloader (or other loader tool) problem, and have TABs contain only unique TBFs but allow tockloader to use a more generic arch if the specific arch doesn't exist.

alexandruradovici: I think a) is easier for now. I'll revert the changes. In the long term, I think it might be worth discussing in detail how we should change tockloader to have a better fix than this.

bradjc:
Can you pull the two commits from https://github.com/tock/elf2tab/compare/pr/38?expand=1?

I also tried this with libtock-c and it seems to work fine.

alexandruradovici:
Done

hudson-ayers (approved):

dcz-self:
Is bors actually enabled on this repo? This is taking long, and meanwhile, libtock-rs doesn't produce working TAB files.

hudson-ayers:
oops! thanks for checking back on this. bors is not enabled on this repo

PR #41, opened 2022-05-24
Convert elf2tab into a lib+bin crate

cfrantz:
The rationale for this change is to allow cargo-raze to process this
crate as a dependency.  As a bin-only crate, cargo-raze requires
the crate to be published at crates.io and does not permit using a
git repo as the source.

As a bin+lib crate, there is no such restriction.

Signed-off-by: Chris Frantz <cfrantz@google.com>


hudson-ayers (approved):
Tested that this still builds the binary as expected locally

cfrantz:
Note: This is a known issue.  https://github.com/google/cargo-raze/issues/446

PR #43, opened 2022-06-26
Add support for adding credentials to a userspace binary

phil-levis:
Verifying the integrity of userspace binaries involves adding credentials footers. Adding credentials footers requires adding a header that distinguishes the end of the program binary (executable code) from the end of the userspace binary (the whole binary image including footers). This is necessary because footers are not covered by integrity. 

This pull request:
  1. Adds the `-sha256`, `-sha512`, and `-rsa4096` flags; each flag adds the corresponding footer. The RSA4096 footer uses a PKCS#1 v1.5 signature with SHA512.
  2. Adds support for a Program Header. A Program Header is an extended Main Header, which contains two additional fields. One indicates the offset within the binary at which code ends (and footers begin). The second indicates the version of the userspace binary. The version field allows supporting new versions of the same application. `elf2tab` now always inserts both a Program and a Main Header, the latter for backwards compatibility with older kernels.


src/cmdline.rs
```diff
@@ -64,6 +64,15 @@ pub struct Opt {
     #[structopt(long = "deterministic", help = "Produce a deterministic TAB file")]
     pub deterministic: bool,
 
+    #[structopt(long = "program", help = "Use a program header")]
+    pub program: bool,
+
+    #[structopt(long = "app-version",
+                help = "Set the version number",
+                default_value = "19",
```
alistair23, line 72: Why 19?

phil-levis: I set the default value to 0.

src/cmdline.rs
```diff
@@ -64,6 +64,15 @@ pub struct Opt {
     #[structopt(long = "deterministic", help = "Produce a deterministic TAB file")]
     pub deterministic: bool,
 
+    #[structopt(long = "program", help = "Use a program header")]
+    pub program: bool,
```
bradjc, line 68: What is the intent behind main vs program tlv? I know we need to support both, but does it make sense to just prioritize program header (and use it in all of our tooling) and then drop support for main at some point in the future? Or should we plan to use both, where program supports footers and main is for apps without footers.

In any case, this doesn't seem like a flag we should have, the program header should be used in any case where elf2tab needs to insert a footer.

src/convert.rs
```diff
@@ -494,11 +532,41 @@ pub fn elf_to_tbf<W: Write>(
         }
     }
 
-    // Add the relocation data to our total length. Also include the 4 bytes for
-    // the relocation data length.
+    // Add the relocation data to our total length. Also include the
+    // 4 bytes for the relocation data length.
     binary_index += relocation_binary.len() + mem::size_of::<u32>();
+    
+    if program {
```
bradjc, line 539: This check should be if any footers are requested.

bradjc:
Just FYI this change seems to need a somewhat (?) new cargo. Originally I got:

```
cargo install --path .
  Installing elf2tab v0.10.0-dev (/Users/bradjc/git/elf2tab)
    Updating crates.io index
  Downloaded crypto-common v0.1.6
  Downloaded thiserror v1.0.31
  Downloaded typenum v1.15.0
  Downloaded simple_asn1 v0.6.2
  Downloaded proc-macro2 v1.0.40
  Downloaded num_threads v0.1.6
  Downloaded untrusted v0.7.1
  Downloaded cpufeatures v0.2.2
  Downloaded digest v0.10.3
  Downloaded thiserror-impl v1.0.31
  Downloaded spin v0.5.2
  Downloaded num-integer v0.1.45
  Downloaded num-bigint v0.4.3
  Downloaded time-macros v0.2.4
  Downloaded sha2 v0.10.2
  Downloaded block-buffer v0.10.2
  Downloaded num-traits v0.2.15
  Downloaded generic-array v0.14.5
  Downloaded unicode-ident v1.0.2
  Downloaded itoa v1.0.2
  Downloaded filetime v0.2.17
  Downloaded time v0.3.11
error: failed to compile `elf2tab v0.10.0-dev (/Users/bradjc/git/elf2tab)`, intermediate artifacts can be found at `/Users/bradjc/git/elf2tab/target`

Caused by:
  failed to parse manifest at `/Users/bradjc/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.3.11/Cargo.toml`

Caused by:
  feature `edition2021` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["edition2021"]` to enable this feature
```

I don't think this is a problem, just something to be aware of.

src/header.rs
```diff
@@ -592,10 +721,14 @@ impl fmt::Display for TbfHeader {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "TBF Header:")?;
         write!(f, "{}", self.hdr_base)?;
-        write!(f, "{}", self.hdr_main)?;
+        write!(f, "{:?}", self.hdr_main)?;
```
bradjc, line 724:
```suggestion
                self.hdr_main
            .map_or(Ok(()), |hdr| write!(f, "{}", hdr))?;
```

Otherwise it prints it like:

```
Some(TbfHeaderMain { base: TbfHeaderTlv { tipe: Main, length: 12 }, init_fn_offset: 41, protected_size: 0, minimum_ram_size: 4604 })
```

src/cmdline.rs
phil-levis: I think your two comments are mutually exclusive. If we want to transition to always using a program header, then it can't only be when there's a footer. E.g., it should be possible to have a TBF end at the end of the application binary. Having a footer requires you use the program header, but if you don't have footers you can still have a program header.

I originally argued we should transition to program headers and move away from main headers, and you argued instead we should continue to support both, in part for kernel versioning reasons. So a TBF can have both a main header (for older kernels) and a program header (for newer kernels). You argued that if both are present it should be up to the kernel to decide which one to use.

src/convert.rs
phil-levis: Please see above -- a footer automatically adds a program header, but there are cases when you want a program header but do not have footers.

src/cmdline.rs
bradjc: There are three cases I think:

- If a there is no footer:
    - We must include Main. That supports old kernels.
- If there is a footer:
    - We can include Program and not Main. Arguably don't need to support old kernels, why compile with footer for old kernels.
    - We can include both.

The only case where Program is optional is when there is no footer. In that case, I suppose elf2tab could be explicitly instructed to add a Program tlv using the `--program` flag. However, I don't think that should be an option. Either we have elf2tab always insert a Program header, or elf2tab only uses it when necessary, but we don't expose this as an option, mainly because 1) users shouldn't have to think through this logic, and 2) its confusing what should happen if a footer is specified but _not_ `--program`.

Also, I'm supportive of moving towards only Program, but for backwards compatibility we have to support both.

phil-levis: OK, I will remove the option.

src/convert.rs
phil-levis: Removed.

src/header.rs
phil-levis: Done.

phil-levis:
@bradjc I've removed the `--program` option and have Program Headers always included, as well as address your other comments.

bradjc (approved):

src/convert.rs
```diff
@@ -534,6 +594,167 @@ pub fn elf_to_tbf<W: Write>(
...
+            base: header::TbfHeaderTlv {
+                tipe: header::TbfHeaderTypes::Credentials,
+                length: sha256_tlv_len as u16,
+            },
+            format: header::TbfFooterCredentialsType::SHA256,
+            data: result.to_vec(),
+        };
+        output.write_all(sha_credentials.generate().unwrap().get_ref())?;
+        footer_space_remaining -= sha256_len;
+        println!("Added SHA256 credential.");
```
bradjc, line 623: This (and others) should be gated on verbose mode.

phil-levis: Done.

src/convert.rs
```diff
@@ -534,6 +594,173 @@ pub fn elf_to_tbf<W: Write>(
...
+
+        output.write_all(rsa4096_credentials.generate().unwrap().get_ref())?;
+        footer_space_remaining -= rsa4096_len;
+        if verbose {
+            println!("Added PKCS#1v1.5 RSA4096 signature credential.");
+        }
+    }
+
+    let padding_len = footer_space_remaining;
+    if padding_len >= mem::size_of::<header::TbfHeaderTlv>() {
```
bradjc, line 750: I think this check needs to be `if padding_len >= mem::size_of::<header::TbfFooterCredentials>() {`. Otherwise `padding_len` could be 4 yet this would add an 8 byte TLV.

bradjc:
Might be helpful to document how to get an rsa key (what I did):

```
$ openssl genrsa -aes256 -out tockkey.private.pem 4096
$ openssl rsa -in tockkey.private.pem -outform der -out tockkey.private.der
$ openssl rsa -in tockkey.private.pem -outform der -pubout -out tockkey.public.der
```

phil-levis:
> Might be helpful to document how to get an rsa key (what I did):
> 
> ```
> $ openssl genrsa -aes256 -out tockkey.private.pem 4096
> $ openssl rsa -in tockkey.private.pem -outform der -out tockkey.private.der
> $ openssl rsa -in tockkey.private.pem -outform der -pubout -out tockkey.public.der
> ```

Yes, great point. What's the right place to document this?

bradjc:
I added it to the elf2tab readme as well as more credential documentation.

I also added SHA384 support.

phil-levis:
OK - look like this is good to go? Let's discuss at the core meeting today.

README.md
```diff
@@ -17,22 +17,29 @@ Converts Tock userspace programs from .elf files to Tock Application Bundles.
...
         --minimum-ram-size <min-ram-size>                  in bytes
+        --minimum-footer-size <min-footer-size>            Minimum number of bytes to reserve space for in the footer [default: 0]
         --permissions <permissions>...                     A list of driver numbers and allowed commands
     -n, --package-name <pkg-name>                          package name
         --protected-region-size <protected-region-size>    Size of the protected region (including headers)
         --read_ids <read_ids>...                           Storage IDs that this app is allowed to read
-        --stack <stack-size>                               in bytes [default: 2048]
+        --rsa4096-private <rsa4096-private-key>            Add an 4096-bit RSA signature credential using this private key
+        --rsa4096-public <rsa4096-public-key>              Add an 4096-bit RSA signature credential containing this public key
+        --stack <stack-size>                               in bytes
```
phil-levis, line 42: It looks like a change by @bradjc  -- I'll defer to his judgement here.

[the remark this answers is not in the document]
bradjc: I think that was manually added? I think I copied from `elf2tab --help`.

bradjc:
I updated this to current master and added some new changes.

We should do another round of testing with the kernel PR, but I think we should merge this soon since we have the 1.10 release that most people can use.

PR #44, opened 2022-07-26
tbf: persistent ACL: fix length calculation

bradjc:
Each read ID and access ID is four bytes.

@alistair23


alistair23 (approved):

hudson-ayers (approved):
bors r+

Issue #45, opened 2022-08-10
When padding is added to the output tbf file, the init_fn_offset is not calculated correctly


nabieasaurus:
We are working on the x86 port of the elf2tab and discovered this bug in the process.

The command line parameter for the tool is
`elf2tab -n hello --stack 2048 --app-heap 1024 --kernel-heap 1024 --kernel-major 2 --kernel-minor 0 -v -o hello.tab i686.elf`

Output
```

Creating "i686.tbf"
Min RAM size from segments in ELF: 632 bytes
Number of writeable flash regions: 0
Kernel version: 2.0
  Adding .crt0_header section. Offset: 52 (0x34). Length: 40 (0x28) bytes.
Entry point is in .text section
  Adding 24 bytes of padding between sections
  Adding .text section. Offset: 116 (0x74). Length: 896 (0x380) bytes.
  Adding .got section. Offset: 1012 (0x3f4). Length: 12 (0xc) bytes.
  Adding 20 bytes of padding between sections
  Adding .data section. Offset: 1044 (0x414). Length: 128 (0x80) bytes.
Searching for .rel.X sections to add.
TBF Header:
               version:        2        0x2
           header_size:       52       0x34
            total_size:     1176      0x498
                 flags:        1        0x1

        init_fn_offset:       40       0x28
        protected_size:        0        0x0
      minimum_ram_size:     4728     0x1278

    kernel version: ^2.0

Application size report for target i686:
i686-linux-gnu-size i686.elf
   text    data     bss     dec     hex filename
    977     308    2412    3697     e71 i686.elf
make: Leaving directory '/osse.fw/sources/apps/hello'
[cargo-make] INFO - Build Done in 7.44 seconds.
```

The `init_fn_offset` should be 40 + 24 (padding) == 64 but the tool conveniently skips the padding which makes the start point to the start of padding instead of the actual start point.

Attached is the elf file that is given as input.
[i686.zip](https://github.com/tock/elf2tab/files/9303718/i686.zip)

PR #46, opened 2022-08-24
fix entry point calculation

reynoldsbd:
Insertion of padding causes binary_index to be incremented. We must wait to compute entry point until after this happens.

Addresses #45 (cc @nabieasaurus)

@bradjc - Adding you here as it looks like you are the one who originally introduced the padding logic.

This PR basically just moves entrypoint calculation to _after_ padding insertion. This also results in the logic being moved _into_ an `if` block, but semantically I think it makes sense (we should be able to assume that the section containing entrypoint is always nonzero/executable/progbits)

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -362,27 +362,6 @@ pub fn elf_to_tbf<W: Write>(
     for s in &sections_sort {
         let section = &input.sections[s.0];
 
-        // Determine if this is the section where the entry point is in. If it
-        // is, then we need to calculate the correct init_fn_offset.
-        if input.ehdr.entry >= section.shdr.addr
-            && input.ehdr.entry < (section.shdr.addr + section.shdr.size)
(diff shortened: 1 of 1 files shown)


bradjc (approved):
I agree and this change looks correct. I think we've always had the entry point in the first section, so we've never hit this case before.

That TODO on line 352 looms larger and larger :)

hudson-ayers (approved):

reynoldsbd:
Thanks for the approvals!

Are there any plans to publish a new version to crates.io? If not it's no big deal, I can just use `cargo install` with `--git`

hudson-ayers:
I would guess we will not have another release until https://github.com/tock/elf2tab/pull/43 is merged

reynoldsbd:
Anything you need from me before this PR can be merged?

PR #48, opened 2022-09-08
Fix fixed flash address calculation

bradjc:
If a segment does not have a .start section we still must calculate the correct first flash address.

I don't think libtock-c uses `.start` which is why #39 doesn't quite work.

This is a fix, but it might not be what we want. Should we look for _any_ section in the segment?

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -149,9 +149,12 @@ pub fn elf_to_tbf<W: Write>(
     /// Helper function to find the start section is inside a
     /// given segment.
     ///
-    /// This is necessary because the flash segment is not guaranteed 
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
I think this looks good. I tried to reason through any scenario where not filtering for the .start section would be a problem and could not think of one. I assume that if a segment is an executable segment, all sections in it should be executable?

hudson-ayers:
Whoops, I had both PRs open but meant to approve the other one

hudson-ayers:
In general, it seems if there is an executable section before `.start` we still want that to end up in the binary, so #49 is better imo

PR #50, opened 2022-09-09
Add CI for rustfmt

bradjc:
My attempt to add a check for rustfmt to the CI.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,8 +1,6 @@
 # This workflow checks that the code is formatted using rustfmt.
 
 name: tock-ci
-env:
-  TERM: xterm # Makes tput work in actions output
 
 # Controls when the action will run. Triggers the workflow on push or pull
 # request events but only for the master branch.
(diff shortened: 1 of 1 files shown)


.github/workflows/format.yml
```diff
@@ -0,0 +1,33 @@
+# This workflow checks that the code is formatted using rustfmt.
+
+name: tock-ci
+env:
+  TERM: xterm # Makes tput work in actions output
```
hudson-ayers, line 5: We don't need this, it was just in the Tock workflow because we use tput in the Makefiles, but it doesn't hurt to leave it here

.github/workflows/format.yml
```diff
@@ -0,0 +1,33 @@
...
+      matrix:
+        os: [ubuntu-latest]
+    # The type of runner that the job will run on
+    runs-on: ${{ matrix.os }}
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1 # pulls version from rust-toolchain file
+      - uses: actions/setup-node@v1
```
hudson-ayers, line 29: The Tock workflow installs node for the markdown checks it does, but that just slows us down here

hudson-ayers:
You'll need to actually run rustfmt on the repo for this to go through

hudson-ayers:
I think you need to rebase for this PR to pass the "push" workflow

hudson-ayers (approved):

PR #58, opened 2022-10-04
Uses program headers to create TBF File

bincsh:
Replaces the section iterator to create the TBF data in favor of iterating over program headers and using the program segments structure.

Tested this change on all libtock-c examples compiled for Risc-V and ARM. From 1716 ELFs, using elf2tab/master and elf2tab/PR produce all 1716 TBFs with the same file size. 1704 TBF were binary identical. The remaining 12 TBFs, elf2tab/PR adds the missing unwinding information (from .ARM.exidx, present in the program headers), which elf2tab/master doesn't.

Using program headers is the method most (if not all) ELF loaders utilize when loading ELFs in memory.  ~~Elf2Tock will look for .crt0_header section, which is the first expected data after the TBF header in memory. This is necessary since some linkers are mapping the entire ELF file into memory (i.e. program header offset = 0). The code verifies this corner case and adjusts the offset within the initial program segment to ensure .crt0_header section content is the first thing after TBF headers.~~


src/convert.rs
```diff
@@ -411,6 +421,28 @@ pub fn elf_to_tbf(
...
+        .sections
+        .iter()
+        .find_map(|section| {
+            if section.shdr.name == ".crt0_header" {
+                Some(section.shdr.addr as u64)
+            } else {
+                None
+            }
+        })
+        .expect("ELF doesn't have .crt0_header section");
```
bradjc, line 444: We don't want this, as we don't want a requirement that there be a section named `.crt0_header`. That is just a convention that libtock-c uses, but not something we want elf2tab to rely on.

bradjc:
> This is necessary since some linkers are mapping the entire ELF file into memory (i.e. program header offset = 0). The code verifies this corner case and adjusts the offset within the initial program segment to ensure .crt0_header section content is the first thing after TBF headers.

Can we dig into this more? What do these ELFs look like? This sounds like an issue on the linking side that we need to investigate.

src/main.rs
```diff
@@ -146,7 +138,6 @@ fn main() {
             opt.permissions.to_vec(),
             (opt.write_id, opt.read_ids.clone(), opt.access_ids.clone()),
             minimum_tock_kernel_version,
-            add_trailing_padding,
```
bradjc, line 149: We had this as an argument because we don't necessarily _want_ to include this padding, but we do because it makes using TABs on cortex-m easier. In the future we may want to change when we add trailing padding. I think we should keep this argument.

bincsh: I moved this argument to inside elf_to_tbf function, since elf_file is moved to convert.rs:56~63. This argument (add_trailing_padding) is taken from elf::File.

src/convert.rs
bincsh: These are the files that fell into this category:
examples/tests/bare_bones/rv32imc/rv32imc.0x20030880.0x10008000
examples/tests/bare_bones/rv32imc/rv32imc.0x20032080.0x10008000 
examples/tests/bare_bones/rv32imc/rv32imc.0x20034080.0x10008000
examples/tests/bare_bones/rv32imc/rv32imc.0x20030080.0x10005000
examples/tests/crash_immediately_syscall/rv32imc/rv32imc.0x20030880.0x10008000 
examples/tests/crash_immediately_syscall/rv32imc/rv32imc.0x20032080.0x10008000
examples/tests/crash_immediately_syscall/rv32imc/rv32imc.0x20034080.0x10008000
examples/tests/crash_immediately_syscall/rv32imc/rv32imc.0x20030080.0x10005000

This is how one of the program headers look like (others look similar)
Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  LOAD           **0x000000** 0x20030000 0x20030000 0x008b0 0x008b0 RWE 0x1000
  LOAD           0x001000 0x10008000 0x200308b0 0x00000 0x00000 RW  0x1000

I can attach these files if you need them.

hudson-ayers: Notably those are the two tests without a `main()` function

src/convert.rs
```diff
@@ -424,132 +406,111 @@ pub fn elf_to_tbf(
...
+    // memory block. Only take into consideration segments where filesz is greater than 0
+    for segment in &elf_file.phdrs {
+        match segment.progtype {
+            elf::types::PT_LOAD => {
+                if segment.filesz > 0 {
+                    if last_segment_address_end.is_some() {
+                        // We have a previous section. Now, check if there is any
+                        // padding between the sections in the .elf.
+                        let padding = segment.paddr as usize - last_segment_address_end.unwrap();
+                        if padding < 1024 {
```
bradjc, line 424: We shouldn't need this check. Since we using the segments, if the elf tells us to include the segment then we should just do it and insert padding as needed. We don't have to worry about using a section we shouldn't be using like we did before.

src/convert.rs
```diff
@@ -424,132 +406,111 @@ pub fn elf_to_tbf(
...
-                                println!("  Adding {} bytes of padding between sections", padding,);
+    let mut start_address: u64 = 0;
+    let start_program_binary_index = binary_index;
+
+    // Iterate over ELF's Program Headers to assemble the binary image as a contiguous
+    // memory block. Only take into consideration segments where filesz is greater than 0
+    for segment in &elf_file.phdrs {
+        match segment.progtype {
+            elf::types::PT_LOAD => {
+                if segment.filesz > 0 {
```
bradjc, line 419:
```suggestion
                // Do not include segments with zero size, as these likely go in memory, not flash.
                if segment.filesz > 0 {
```

src/convert.rs
```diff
@@ -424,132 +406,111 @@ pub fn elf_to_tbf(
...
+                    if last_segment_address_end.is_some() {
+                        // We have a previous section. Now, check if there is any
+                        // padding between the sections in the .elf.
```
bradjc, line 422:
```suggestion
                        // We have a previous segment. Now, check if there is any
                        // padding between the segments in the .elf.
```

src/convert.rs
```diff
@@ -424,132 +406,111 @@ pub fn elf_to_tbf(
...
+                        let padding = segment.paddr as usize - last_segment_address_end.unwrap();
+                        if padding < 1024 {
+                            if padding > 0 {
+                                if verbose {
+                                    println!("Padding between Program Segments of {}", padding);
+                                }
                             }
```
bradjc, line 429: I don't think we need to have a message about this.

src/convert.rs
```diff
@@ -424,132 +406,111 @@ pub fn elf_to_tbf(
...
+                        start_address = segment.paddr;
+
+                        // check if start of segment matches the fixed address flash (for non-PIC files)
+                        if let Some(fixed_start_address) = fixed_address_flash {
+                            if start_address != fixed_start_address as u64 {
+                                panic!("First Segment address ({:#x}) is different than fixed_address_flash ({:#x})",
+                                    start_address,
+                                    fixed_start_address);
+                            }
+                        }
```
bradjc, line 452: This check seems backwards. We figure out what the fixed flash address is from the elf, so it doesn't make sense to then check if we figured out the fixed flash address correctly.

src/convert.rs
```diff
@@ -284,7 +266,7 @@ pub fn elf_to_tbf(
     let mut writeable_flash_regions_count = 0;
 
     for s in &sections_sort {
```
bradjc, line 268: I think we should be getting rid of this loop. Or, perhaps removing everything not related to writeable flash regions, and moving it to just the wfr handling section down below.

src/convert.rs
```diff
@@ -284,7 +266,7 @@ pub fn elf_to_tbf(
...
-        let section = &input.sections[s.0];
+        let section = &elf_file.sections[s.0];
```
bradjc, line 269: (sorry, github won't let me put the comment where I want)

We should move populating the `rel_sections` array to after the segment loop before. Basically, we no longer need to "guess" which `.rel.X` sections to add, we should know based on which sections were in loaded segments.

src/convert.rs
```diff
@@ -167,37 +176,8 @@ pub fn elf_to_tbf(
...
-                    None => Some(section.shdr.addr as u32),
-                };
-            }
-        }
-        first_section_address
-    }
-
     // Do flash address.
-    for segment in &input.phdrs {
+    for segment in &elf_file.phdrs {
```
bradjc, line 180: Since we are already looping segments below, can we merge the logic in this loop with the one below?

src/convert.rs
bincsh: What's your suggestion? During the program header loop to detect which sections belongs to the segments?

src/convert.rs
bincsh: My understanding is that the first loop is required to fetch some address in order to build the tbf header. I don't think we could make it in a single loop. Can you clarify?

src/convert.rs
bincsh: Could you please elaborate it further? Is this only about the rel_sections to populate the sections?

bradjc:
This is pretty close. I'm currently working on a few updates. We should be able to get rid of all of the hacky stuff we have been doing to cobble together TBF binaries.

bradjc:
I added a commit which I think gets this PR to the point where it fully removes the kludgy ELF parsing and replaces it with something more principled based on the actual ELF files

One note, @dcz-self, this removes the "check for the first section in the segment" logic when determining addresses. Since we are now including entire segments and not on a section-by-section basis, it doesn't make sense to calculate from the middle of a segment. The TBF will now contain any padding data at the beginning of the segment. I think this is a more principled approach.

bincsh:
> 

Your commit looks good to me. Is there anything else missing?

jrvanwhy:
I gave this a try with `libtock-rs`'s examples on an emulated HiFive and it seems to work fine :-)

PR #60, opened 2022-10-21
Update elf crate to v0.7.0

cole14:
The changes from 0.0.10 -> 0.7.0 represent implementing more complete ELF parsing
functionality, performance improvements for zero-alloc and zero-copy parsing methods,
fuzz testing of the interface, and various bug fixes. v0.7.0 marks the stabilized
release for these changes - i.e. there are no currently planned additional changes
to the library % fixes for any bugs that get reported.

The changes seen in this patch aim to fix it up for the new interface.

The elf crate now supports two parsing modes:
1. Parsing lazily out of a &[u8] of the whole file data (the route I chose here)
2. Parsing lazily out of a std::io::<Read + Seek> via elf::ElfStream

It seemed to me like this tool inspects a lot of the file data anyway, so the former
is likely faster as opposed to the latter which allows lazy i/o as the parser seeks
around the file parsing things as requested.

This patch successfully built locally and the few unit tests in the crate passed, though there doesn't appear to be coverage of this convert.rs code so I'm not sure what the best way is to verify that this didn't accidentally break something in the actual TAB output.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -1,29 +1,119 @@
+//! Convert ELF to TBF.
+
 use crate::header;
 use crate::util::{self, align_to, amount_alignment_needed};
 use ring::{rand, signature};
 use rsa_der;
 use sha2::{Digest, Sha256, Sha384, Sha512};
 use std::cmp;
-use std::io;
-use std::io::Write;
+use std::io::{Read, Seek, SeekFrom, Write};
 use std::mem;
 use std::path::{Path, PathBuf};
+use std::{fs, io};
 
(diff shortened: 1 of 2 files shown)


alistair23 (approved):

bradjc:
Once we decide on #58 I will rebase this on top of those changes.

cole14:
I'm happy to help with that as well if needed

cole14:
Cool, I see #58 was merged. I'll take a look at redoing this with those changes next week

PR #63, opened 2022-12-14
Fix time CVE

kaczmarczyck:
Removes the optional and unused time dependency. `cargo audit` reported a CVE for this crate, but it is not used.

The required minimal `Cargo.lock` changes have been generated with `cargo update -p chrono`.

To verify, install audit with `cargo install cargo-audit` and then run `cargo audit` with and without this commit, the difference is:

```
Crate:     time
Version:   0.1.44
Title:     Potential segfault in the time crate
Date:      2020-11-18
ID:        RUSTSEC-2020-0071
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:  Upgrade to >=0.2.23
Dependency tree:
time 0.1.44
└── chrono 0.4.22
    └── elf2tab 0.11.0-dev

error: 1 vulnerability found!
```

If this fix works for you, I wouldn't mind a new (maybe even backported) release. :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,16 +69,13 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.22"
+version = "0.4.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bfd4d1b31faaa3a89d7934dbded3111da0d2ef28e3ebccdb4f0179f5929d1ef1"
+checksum = "16b0a3d9ed01224b22057780a37bb8c5dbfe1be8ba48678e7bf57ec4b385411f"
 dependencies = [
  "iana-time-zone",
- "js-sys",
  "num-integer",
  "num-traits",
- "time 0.1.44",
- "wasm-bindgen",
  "winapi",
 ]
 
@@ -451,7 +448,7 @@ dependencies = [
  "num-bigint",
  "num-traits",
  "thiserror",
- "time 0.3.14",
+ "time",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


jrvanwhy (approved):

brghena (approved):

hudson-ayers (approved):

Cargo.toml
```diff
@@ -14,7 +14,7 @@ name = "elf2tab"
 path = "src/main.rs"
 
 [dependencies]
-chrono = "0.4.2"
+chrono = { version = "0.4.2", default-features = false, features = ["clock", "std"] }
```
alistair23, line 17: Why not just update chrono to the latest version instead?

alistair23 (changes requested):

Cargo.toml
kaczmarczyck: The [latest `chrono`](https://crates.io/crates/chrono/0.4.23/dependencies) still has the optional `time` dependency that cargo audit complains about.

alistair23: Interesting, it seems like Chrono devs think it [isn't vunrable](https://github.com/chronotope/chrono/issues/602) and will remove the `oldtime` feature in the next version as well

alistair23 (approved):

Cargo.lock
```diff
@@ -69,16 +69,13 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.22"
+version = "0.4.23"
```
bradjc, line 72: So we ask for version "0.4.2" but now we are using version "0.4.23"? Is that what this means? So we are using the latest version of chrono?

kaczmarczyck: Indeed, but it is not an important part of the PR. `cargo update` just happened to also update `chrono` to the latest version. The important information in this file is that `time` is removed, as it is a version that `cargo audit` reports as vulnerable.

lschuermann: @bradjc This boils down to `Cargo.toml` dependency specifications by default assuming SemVer-compatible versioning schemes: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

If we didn't have a `Cargo.lock` file, a user building `elf2tab` would automatically get the latest version of `chrono` within the constraints of `[0.4.2, 0.5]`. IMHO this is anything but intuitive (other tools require an explicit `~>` or `^`), and emphasizes the importance of the lock file.

lschuermann (approved):

kaczmarczyck:
Hi, anything I have to do to help merge this?

PR #64, opened 2023-01-09
Update elf crate to v0.7.1

cole14:
Note: This is a rework of #60 for the new segment-based conversion logic.

The changes from 0.0.10 -> 0.7.1 represent implementing more complete ELF parsing
functionality, performance improvements for zero-alloc and zero-copy parsing methods,
fuzz testing of the interface, and various bug fixes. v0.7.1 marks the stabilized
release for these changes - i.e. there are no currently planned additional changes
to the library % fixes for any bugs that get reported.

The changes seen in this patch aim to fix it up for the new interface.

The elf crate now supports two parsing modes:

    Parsing lazily out of a &[u8] of the whole file data (the route I chose here)
    Parsing lazily out of a std::io::<Read + Seek> via elf::ElfStream

It seemed to me like this tool inspects a lot of the file data anyway, so the former
is likely faster as opposed to the latter which allows lazy i/o as the parser seeks
around the file parsing things as requested.

This patch successfully built locally and the few unit tests in the crate passed, though there doesn't appear to be coverage of this convert.rs code so I'm not sure what the best way is to verify that this didn't accidentally break something in the actual TAB output.

It seemed to me like this tool inspects a lot of the file data anyway, so the former is likely faster as opposed to the latter which allows lazy i/o as the parser seeks around the file parsing things as requested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,12 +50,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c1ad822118d20d2c234f427000d5acc36eabe1e29a348c89b63dd60b13f28e5d"
 
 [[package]]
-name = "byteorder"
-version = "0.5.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


bradjc (approved):
Thanks again for doing this update!!

Issue #56, opened 2022-09-28
Tracking: Rewrite section selection and placement logic


bradjc:
`elf2tab` currently effectively uses a series of heuristics to choose which sections from an ELF to include in the TBF. These heuristics are based on a series of assumptions about 1) how the compilation toolchain created the ELF, 2) what types of sections are in Tock app binaries, and 3) what types of alignments and offsets will be present in the ELF. When elf2tab started, these assumptions were fairly strong (we only had a small number of hardware platforms, a single architecture, and a single toolchain), and therefore it was feasible to code the heuristics to get working TBF files.

Over time, however, Tock has expanded and many of the assumptions are no longer so narrow. We've been tweaking our heuristics to get working TBFs for a much wider set of input ELFs, but it is probably becoming increasingly untenable that our heuristics will continue to work across all platforms and toolchains.

This tracking issue is for a rewrite of how we pull sections from an ELF and include them in a TBF. If I'm not mistaken, we should be able to use the structure of the ELF format, and not our homegrown heuristics, to more universally generate TBFs.

At a high level, my understanding is that we should be:

1. Looping over ELF segments, identifying if those segments need to be in the TBF, and then choosing sections based on which sections are in those segments.
2. Respecting the padding and alignment constraints of segments and sections.
3. Keeping some method of handling PIC apps versus fixed address apps.

This doesn't need to necessarily be tied to this effort, but we do at some point need to think about maintaining alignment constraints with fixed address apps, as noted here: https://github.com/tock/tock/issues/3203. The issue here is that we don't have a way to tell the loader how it needs to align the overall TBF such that the actual code sections fall at the correct alignment.

PR #66, opened 2023-02-14
Fix relocation data offset in TBF header

alexandruradovici:
This pull request fixes the offset of the relocation section in the TBF header.

It seems that starting with version 0.10.0, `elf2tab` has a different way of setting up the TBF file from the ELF sections. Before this version, the ARM sections, like `.ARM.exidx` where ignored. They seem to be added now exactly where the relocation section should be. This make apps that include this section to fault. Not all apps have include this section, for instance a simple *hello* does not.

This is how the sections look like:
```
Min RAM size from segments in ELF: 13012 bytes
Number of writeable flash regions: 0
Kernel version: 2.0
  Adding segment. Offset: 72 (0x48). Length: 162972 (0x27c9c) bytes.
    Contains section .crt0_header. Offset: 72 (0x48). Length: 40 (0x28) bytes.
    Contains section .text. Offset: 112 (0x70). Length: 162932 (0x27c74) bytes.
  Adding segment. Offset: 163044 (0x27ce4). Length: 3336 (0xd08) bytes.
    Contains section .got. Offset: 163044 (0x27ce4). Length: 1324 (0x52c) bytes.
    Contains section .data. Offset: 164368 (0x28210). Length: 2012 (0x7dc) bytes.
      Including relocation data (.rel.data). Length: 2488 (0x9b8) bytes.
    Contains section .bss. Offset: 166380 (0x289ec). Length: 9676 (0x25cc) bytes.
  Adding segment. Offset: 166380 (0x289ec). Length: 8 (0x8) bytes.
    Contains section .ARM.exidx. Offset: 166380 (0x289ec). Length: 8 (0x8) bytes.
TBF Header:
               version:        2        0x2
           header_size:       72       0x48
            total_size:   262144    0x40000
                 flags:        1        0x1

        init_fn_offset:       41       0x29
        protected_size:        0        0x0
      minimum_ram_size:    58132     0xE314

        init_fn_offset:       41       0x29
        protected_size:        0        0x0
      minimum_ram_size:    58132     0xE314
     binary_end_offset:   168880    0x293B0
           app_version:        0        0x0

    kernel version: ^2.0
```

I would need this pull request merged for the Rust Nation presentation.

I am not sure that this is the best fix, but it seems to be the fastest. It is interesting that elf2tab now includes a `bss` section. Shouldn't this be excluded?

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -754,8 +754,14 @@ pub fn elf_to_tbf(
         print!("{}", tbfheader);
     }
 
-    // Write the header and actual app to a binary file.
+    // Write the header to a binary file.
     output.write_all(tbfheader.generate().unwrap().get_ref())?;
+
+    // Adjust the relocation data due to the .ARM.exidx
+    let reldata_start: u32 = binary.len() as u32;
+    binary.as_mut_slice()[32..36].copy_from_slice(&reldata_start.to_le_bytes());
+
+    // Write the actual app to a binary file.
     output.write_all(binary.as_ref())?;
 
     let rel_data_len: [u8; 4] = (relocation_binary.len() as u32).to_le_bytes();


hudson-ayers:
I don't think we want sections like .ARM.exidx in the TAB at all, so I am not sure this is the right fix. libtock-c's linker file explicitly states that it should not be serialized into the binary by elf2tab

hudson-ayers:
Hmm, the description of https://github.com/tock/elf2tab/pull/58 mentions that it adds the unwinding information on purpose, but that is at odds with https://github.com/tock/libtock-c/blob/master/userland_generic.ld#L182-L184

alexandruradovici:
> I don't think we want sections like .ARM.exidx in the TAB at all, so I am not sure this is the right fix. libtock-c's linker file explicitly states that it should not be serialized into the binary by elf2tab

I tried excluding thats section, but that might not be as trivial. We need to skip the segment that contains that section and the only way that I know how to do it is to iterate all sections before starting to add the segment. This is just at a quick overview, it may be easier.

hudson-ayers:
Looks like this is all pretty relevant to https://github.com/tock/elf2tab/issues/56 , in which Brad suggests that ultimately we should be iterating all the sections -- but I realize that is not practical in the short term. 

With the caveat that I am not very familiar with ~any of this, here is what I am thinking:
- whether .ARM.exidx ends up in the TAB should probably not be up to elf2tab making some explicit decision about that section, but up to the linker script used to link apps. E.g. if we do not want those sections ending up in flash, but want them in the elf (for debugging or whatever) we should probably just be marking them as `NOLOAD`
- it is possible that our logic for determining the relocation data offset should be different, but I don't feel comfortable enough with how elf2tab works to make that call. It seems like the fix in this PR only matters if someone has an elf file with a section (.ARM.exidx in our case) loaded into FLASH, after the .endflash symbol. So hopefully the first fix is sufficient.

Does using NOLOAD on the arm exception sections work for you in the short term?

alexandruradovici:
> Does using NOLOAD on the arm exception sections work for you in the short term?

Yes, this works, thank you. I submitted a PR to libtock-c https://github.com/tock/libtock-c/pull/322

ppannuto:
This is really part of a deeper question w.r.t the intent of TABs. If they are to be a general-purpose mechanism for distributing 'board-independent' images of applications, then we need to decide if these application images are solely the 'optimized, to-be-flashed' images or 'all information about' images. The `exidx` section is one of many (optional) debugging sections included in ELFs that are useful for debugging but unused during (normal) runtime. On a traditional system, appending debug info to the end of an application image is no big deal — on an embedded system with limited flash, separating this ( e.g. https://github.com/GabrielMajeri/separate-symbols ) is meritorious.

AFAICT, TABs have no 'debug' story right now (the trivial solution would be to just include the source elf, with all its debug info, in the TAB as well). Fixing that is a larger issue; marking 'non-execution' sections as `NOLOAD` so they don't get pulled into the binary image seems a fine solution for the short term.

hudson-ayers:
> This is really part of a deeper question w.r.t the intent of TABs. If they are to be a general-purpose mechanism for distributing 'board-independent' images of applications, then we need to decide if these application images are solely the 'optimized, to-be-flashed' images or 'all information about' images. The `exidx` section is one of many (optional) debugging sections included in ELFs that are useful for debugging but unused during (normal) runtime. On a traditional system, appending debug info to the end of an application image is no big deal — on an embedded system with limited flash, separating this ( e.g. https://github.com/GabrielMajeri/separate-symbols ) is meritorious.

I do think `exidx` is a little different from normal debugging sections in that on some (slightly more resourceful) embedded systems, it would be possible for the kernel or application runtime to display a backtrace on device using the information in this section. But neither Tock nor any of its userspace runtimes support this, so we are certainly not using the information.

> AFAICT, TABs have no 'debug' story right now (the trivial solution would be to just include the source elf, with all its debug info, in the TAB as well). Fixing that is a larger issue; marking 'non-execution' sections as `NOLOAD` so they don't get pulled into the binary image seems a fine solution for the short term.

+1

hudson-ayers:
Closing this because https://github.com/tock/libtock-c/pull/322 fixes the high-level issue and I am not sure that we would want this change now that it has been resolved

PR #67, opened 2023-04-10
Add padding to 4K align x86 apps

HMiyaziwala:
This code change pads binaries to be 4K aligned for x86 applications so that allocated memory regions can be more correctly handled by an x86 MPU. x86 architectures protect memory at a 4K page granularity, and so an existing x86 MPU would have to grant an application memory outside of its binary if the app is not 4K aligned. 

The change follows the pattern established by the ARM "power of two padding". This PR will affect all ELFs that have an x86 machine value.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -147,13 +147,13 @@ pub fn elf_to_tbf(
     // Load and parse ELF.
     let elf_file = elf::File::open_stream(input_file).expect("Could not open the .elf file.");
 
-    // Adding padding to the end of cortex-m apps. Check for a cortex-m app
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):

bradjc:
Please take a look at #68 and double check it meets your needs.

Since we have two different padding types now it is probably worth making it a little more clear how padding is handled.

HMiyaziwala:
> Please take a look at #68 and double check it meets your needs.
> 
> Since we have two different padding types now it is probably worth making it a little more clear how padding is handled.

I have checked and tested the linked PR, it does what we need. I appreciate the refactor!

PR #69, opened 2023-06-27
Generate segment padding from ELF-internal offsets

lschuermann:
### Pull Request Overview

Previously, `segment.p_addr` was used to generate offsets. Given a specific ordering of ELF segments and certain target memory layouts, this has been observed to insert very large paddings (~256MB), which further break internal offsets in the binary. Instead of the target's addresses, we should be using ELF-file internal offsets to infer padding between segments (which means that the TBF-internal offsets should be equivalent to those of the ELF file itself).

### Testing Strategy

I've tested this with a `libtock-rs` application using a bunch of different linker configurations. Throughout all of these configurations, with this change, `elf2tab` seems to produce binaries that matched the ELF's contents (although some of these apps were still broken for unrelated reasons). Tested with a few `libtock-c` apps as well, although without any linker script changes.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -549,7 +549,9 @@ pub fn elf_to_tbf(
         if let Some(last_segment_address_end) = last_segment_address_end {
             // We have a previous segment. Now, check if there is any padding
             // between the segments in the .elf.
-            let padding = segment.p_paddr as usize - last_segment_address_end;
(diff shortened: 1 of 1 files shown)


bradjc (approved):
Eventually we will get this elf parsing correct!

Nice fix.

PR #70, opened 2023-06-27
Add support for `tbf_protected_region_size` ELF-file symbol

lschuermann:
### Pull Request Overview

This change causes elf2tab determine the TBF protected region size from a special `tbf_protected_region_size` symbol in the ELF file. This can be overriden through the `--protected-region-size` command line parameter. If neither the parameter nor the symbol are present, elf2tab falls back onto its heuristics for determining a reasonable protected region size.

#### Motivation

For non-PIC apps, it is very inconvenient to need to manually calculate the apps' actual text-section start address, and then choose a protected region parameter on elf2tab's command line which matches the pre-calculated flash address such that the TBF is filled up with headers and padding to match the expected app load address. While many times simply specifying an offset in the link address and allowing elf2tab to fall back onto a reasonable well-aligned address, this mechanism can break under some configurations. The following outlines how a PIC and non-PIC target is defined in `libtock-c`:

```
TOCK_TARGETS ?= cortex-m0\
                [...]
                rv32imac|rv32imac.0x20040060.0x80002800|0x20040060|0x80002800\
```

(Here, the intended load address for the application is `0x20040000`. To generate correct binaries, `0x60` is added as an offset to the linker-provided flash address. Deverlopers either rely on elf2tab automatically falling back onto the next-lower 256-byte aligned address, or must pass `--protected-region-size 0x60` to elf2tab.)

`libtock-rs` currently takes a different approach[1]: it reserves a section for the TBF headers at the start of its flash region. However, this does not seem to work reliably. For instance, the linker is free to re-arrange segments in the ELF file, and e.g., have the RAM segment located before the flash segment. While TBF is compatible with that (by setting the `init_fn_offset` field properly), having the TBF header reservation at the front of the `FLASH` segment (which is not the first segment in the ELF file) would break the assumption that the TBF header precedes the actual process binary. Furthermore, the current ELF parsing logic interprets this `.tbf_header` section as the start of application flash and prepends TBF headers before that section.

By introducing support for a `tbf_protected_region_size` symbol (which can be set to an arbitrary value outside of any ELF segment), applications can place their `FLASH` segment at an appropriate offset and hint this offset to `elf2tab`, without placing any constraints on the ELF segment layout itself.

[1]: https://github.com/tock/libtock-rs/blob/0f7c97627b7d49dd34129d40717eadba9d307a2d/runtime/libtock_layout.ld#L50-L54

### Testing Strategy

These changes are tested with a preliminary integration into the linker-scripts of the `libtock-c` and `libtock-rs` toolchains.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -423,52 +423,84 @@ pub fn elf_to_tbf(
     // Adjust the protected region size to make fixed address work
     ////////////////////////////////////////////////////////////////////////////
 
-    // If a protected region size was passed, confirm the header will fit.
-    // Otherwise, use the header size as the protected region size.
+    // Applications can hint a desired protected region size to elf2tab by
+    // defining a special `tbf_protected_region_size` symbol:
+    let protected_region_size_symbol =
+        if let Ok(Some((symtab, sym_strtab))) = elf_file.symbol_table() {
+            // We are looking for the `tbf_protected_region_size` symbol and its
+            // value. If it exists, we can use it as a hint for the protected
+            // region size.
+            symtab
+                .iter()
+                .find(|sym| {
+                    let name = sym_strtab
+                        .get(sym.st_name as usize)
+                        .expect("Failed to parse symbol name");
+                    name == "tbf_protected_region_size"
+                })
+                .map(|tbf_header_sym| tbf_header_sym.st_value as u32)
+        } else {
+            None
+        };
+
+    // Determine the protected region size by checking the following sources in
+    // this order:
+    //
+    // 1. Check for a `tbf_protected_region_size` symbol in the ELF file.
+    //
+    // 2. Use a fixed protected region size if one was passed through a command
+    //    line argument.
+    //
+    // 3. Set the protected region size to fit the TBF headers. For non-PIC
+    //    apps, align the start of the generated TBF file on a 256-byte
+    //    boundary, based on the binary's fixed flash address.
     let protected_region_size =
-        if let Some(fixed_protected_region_size) = protected_region_size_arg {
-            if fixed_protected_region_size < header_length as u32 {
-                // The header doesn't fit in the provided protected region size;
-                // throw an error.
-                return Err(io::Error::new(
-                    io::ErrorKind::InvalidInput,
-                    format!(
-                    "protected_region_size = {} is too small for the TBF headers. Header size: {}",
-                    fixed_protected_region_size, header_length),
-                ));
+        if let Some(fixed_protected_region_size) = protected_region_size_symbol {
+            // The protected region size was specified in the ELF file through
+            // the special `tbf_protected_region_size` symbol.
+            //
+            // If we have also been passed a fixed protected region size on the
+            // command line, warn that the ELF symbol will take precedence!
+            if protected_region_size_symbol.is_some() {
+                println!(
+                    "Overriding command-line specified protected_region_size \
+		 with tbf_protected_region_size symbol = {} bytes!",
+                    fixed_protected_region_size
+                );
             }
-            // Update the header's protected size, as the protected region may
-            // be larger than the header size.
-            tbfheader.set_protected_size(fixed_protected_region_size - header_length as u32);
 
             fixed_protected_region_size
+        } else if let Some(fixed_protected_region_size) = protected_region_size_arg {
+            // A desired protected region size was specified on the command line:
+            fixed_protected_region_size
         } else {
-            // The protected region was not specified on the command line.
-            // Normally, we default to an additional size of 0 for the protected
-            // region beyond the header. However, if we are _not_ doing PIC, we
-            // might want to choose a nonzero sized protected region. Without
-            // PIC, the application binary must be at specific address. In
-            // addition, boards have a fixed address where they start looking
-            // for applications. To make both of those addresses match, we can
-            // expand the protected region.
+            // The protected region was neither specified on the command line,
+            // nor as part of the ELF file. Normally, we default to an
+            // additional size of 0 for the protected region beyond the
+            // header. However, if we are _not_ doing PIC (as enforced in the
+            // check above), we might want to choose a nonzero sized protected
+            // region. Without PIC, the application binary must be at specific
+            // address. In addition, boards have a fixed address where they
+            // start looking for applications. To make both of those addresses
+            // match, we can expand the protected region.
             //
-            // |----------------------|-------------------|---------------------
-            // | TBF Header           | Protected Region  | Application Binary
-            // |----------------------|-------------------|---------------------
-            // ^                                ^         ^
-            // |                                |         |-- Fixed binary address
-            // |-- Start of apps address        |-- Flexible size
+            // /----- Protected Region ----------------\
+            // |------------|------------------------- |---------------------
+            // | TBF Header | Protected Region Trailer | Application Binary
+            // |------------|--------------------------|---------------------
+            // ^                           ^           ^
+            // |                           |           |-- Fixed binary address
+            // |-- Start of apps address   |-- Flexible size
             //
-            // However, we don't actually know the start of apps address.
-            // Additionally, an app may be positioned after another app in
-            // flash, and so the start address is actually the start of apps
-            // address plus the size of the first app. Tockloader when it goes
-            // to actually load the app can check for these addresses and expand
-            // the protected region as needed. But, in some cases it is easier
-            // to just be able to flash the TBF directly onto the board without
-            // needing Tockloader. So, we at least try to pick a reasonable
-            // protected size in the non-PIC case to give the TBF a chance of
-            // working as created.
+            //
+            // An app may be positioned after another app in flash, and so the
+            // start address is actually the start of apps address plus the size
+            // of the first app. Tockloader can check for these addresses and
+            // expand the protected region when it loads the app. But, in some
+            // cases it is easier to just be able to flash the TBF directly onto
+            // the board without needing Tockloader. So, we at least try to pick
+            // a reasonable protected size in the non-PIC case to give the TBF a
+            // chance of working as created.
             //
             // So, we put the start address of the TBF header at an alignment of
             // 256 if the application binary is at the expected address.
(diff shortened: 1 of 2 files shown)


src/convert.rs
```diff
@@ -423,80 +423,137 @@ pub fn elf_to_tbf(
...
+            .map(|tbf_header_sym| tbf_header_sym.st_value as u32)
+    } else {
+        None
+    };
+
+    // If a protected region size was passed, confirm the header will fit. Then,
+    // if this is a non-PIC app, fall back onto the `tbf_header` symbol for
+    // inferring the protected region size. If neither are given, use the header
+    // size as the protected region size.
+    let protected_region_size = if let Some(fixed_protected_region_size) = protected_region_size_arg
```
lschuermann, line 452: This formatting looks weird, but it's been automatically adjusted by `rustfmt` and does not indent the if-block contents by a full 8 spaces, I kept it this way. Let me know if I should try change it.

bradjc:
> and must be also passed to `elf2tab` as `--protected-region-size 0x60`.

That's not right, primarily because elf2tab has to support both pic and non-pic elfs, so always setting the protected region size would generate undesirable pic tbfs.

I ran `make V=1 RISCV=1` and here is the call to elf2tab:

```
elf2tab -n blink --stack 2048 --app-heap 1024 --kernel-heap 1024 --kernel-major 2 --kernel-minor 0 -v -o build/blink.tab build/cortex-m0/cortex-m0.elf build/cortex-m3/cortex-m3.elf build/cortex-m4/cortex-m4.elf build/cortex-m7/cortex-m7.elf build/rv32imac/rv32imac.0x20040060.0x80002800.elf build/rv32imac/rv32imac.0x403B0060.0x3FCC0000.elf build/rv32imc/rv32imc.0x41000060.0x42008000.elf build/rv32imc/rv32imc.0x00080060.0x40008000.elf build/rv32imc/rv32imc.0x20030080.0x10005000.elf build/rv32imc/rv32imc.0x20030880.0x10008000.elf build/rv32imc/rv32imc.0x20032080.0x10008000.elf build/rv32imc/rv32imc.0x20034080.0x10008000.elf build/rv32imac/rv32imac.0x40430060.0x80004000.elf build/rv32imac/rv32imac.0x40440060.0x80007000.elf
```

lschuermann:
> That's not right, primarily because elf2tab has to support both pic and non-pic elfs, so always setting the protected region size would generate undesirable pic tbfs.

Sorry, I should've been more clear: this is for the current state-of-practice for non-PIC apps only. We'd eliminate the need for this option with this PR, assuming we set the `tbf_header` symbol correctly.

You're right -- this is not needed for PIC apps, and in practice you can even get non-PIC apps to work, as the 256 byte `elf2tab` alignment magic has a good chance of working: https://github.com/tock/elf2tab/blob/9e1f707b95470f9b9ea8432bc562dd7f95628d40/src/convert.rs#L476-L491

However, this is brittle and broke for me in the past. To reliably generate working apps at the intended offsets for non-PIC platforms with the current(!) `libtock-c` and `elf2tab`, keeping the linker address offset and the `--protected-region-size` argument in sync is required.

bradjc:
The underlying argument in this PR is that aligning the start of a tbf to a multiple of 256 bytes (our current practice) is insufficient. Why is that?

bradjc:
> Sorry, I should've been more clear: this is for the current state-of-practice for non-PIC apps only.

libtock-c and libtock-rs are quite different, and we've (I've) worked extensively on supporting _both_ PIC and non-PIC architectures, and creating TABs which are board independent for libtock-c. This is not the case for libtock-rs, so I think it is worth discussing why libtock-rs cannot just copy libtock-c's lead.

libtock-c has never used the --protected-region-size flag:

```
❯ pwd
/Users/bradjc/git/libtock-c
❯ git grep protected-region-size | wc -l
       0
```

That flag was added to get support for non-pic apps up and working quickly in libtock-rs.

We don't need to use 256 for the alignment. We could choose something based on the size of the app binary.

Since we add a header in the TBF, once the tbf is generated we know exactly where it needs to be placed to make the app binary start at the correct address. So one question is: why is the app being loaded at the wrong location, and how does adding a start symbol help with that?

lschuermann:
> The underlying argument in this PR is that aligning the start of a tbf to a multiple of 256 bytes (our current practice) is insufficient. Why is that?

This is a good question. In general, aligning "downward" onto 256 bytes should cover most cases. I don't know with certainty what went wrong in my case, but here's some cases where the above code would break (if I understand it correctly):
- if the app is linked so some address `0x140` and the app is intended to be loaded at address `0x100`, but the generated header is larger than 64 bytes, the above magic will reserve 64 bytes but the header size will exceed that and prepend more bytes, breaking the intended app load address
- the above mechanism doesn't support headers larger than 256 byte (unlikely?)

lschuermann:
> Since we add a header in the TBF, once the tbf is generated we know exactly where it needs to be placed to make the app binary start at the correct address. So one question is: why is the app being loaded at the wrong location, and how does adding a start symbol help with that?

I think the core of this issue is that applications are linked with the intention of having them be loaded at a certain address. This address is specified in the linker script's `MEMORY` section. Given the way Tock traverses through the list of TBF apps, it's important to have an app built for the exact specified load address (at least in the non-PIC case).

The unintuitive / "tricky" part is that the an app must not actually be linked to this exact address. In the non-PIC case, the app must be linked to an address which leaves enough headroom for `elf2tab` to prepend the header, such that the app's linker-supplied address _minus_ the prepended header size falls onto the intended app load address. This worked in the past through two mechanisms:
- assume that the app is linked for an address which has sufficient headroom to fix the generated TBF headers. Further, assume that the app load address is 256-byte aligned. Also (at least in the current implementation) the headers shouldn't exceed 256 bytes. Then, `elf2tab` chooses a protected region size appropriate for these assumptions using the above mechanism.
- manually supply the difference between the intended load address and the app's link address, and pass that into `--protected-region-size`

`libtock-rs`' linker script is, in theory, much more intuitive: the developer provides the intended app load address (and optionally a custom TBF header size), and `libtock-rs` will try to reserve a section usable for `elf2tab` to insert its header. This PR tries to take the intuition of this mechanism (have the developer only supply the intended address once, and then let the toolchain figure out placement from there), but circument the issues that `libtock-rs` has. Its is also compatible with `libtock-c`, and does not change the PIC case at all.

One could see this change as a UX adjustment: instead of guessing an offset from within `elf2tab` or requiring the user to manually pass an appropriate linker address and protected region size, have the user provide just the intended address to the linker once.

Here's how this could integrate with libtock-rs: https://github.com/lschuermann/libtock-rs/pull/new/otdev/elf2tab-tbf-header-sym
Notice how the protected size no longer needs to be kept in sync with the linker script.

And this shows how it could be integrated into libtock-c: https://github.com/lschuermann/libtock-c/tree/otdev/elf2tab-tbf-header-sym
Developers would now actually specify the intended load address for non-PIC apps, and can optionally specify a custom TBF header size. `elf2tab` can reliably reconstruct this intended load address and the protected region size from the `tbf_header` symbol. PIC apps are entirely unaffected by these changes.

bradjc:
> I think the core of this issue is that applications are linked with the intention of having them be loaded at a certain address. This address is specified in the linker script's `MEMORY` section. Given the way Tock traverses through the list of TBF apps, it's important to have an app built for the exact specified load address (at least in the non-PIC case).
> 
> The unintuitive / "tricky" part is that the an app must not actually be linked to this exact address.

But this is only true for libtock-rs. This is not true for libtock-c. For libtock-c, it is exactly as expected: we specify the address the app binary should be compiled for and the linker makes it so.

> onto the intended app load address

I think this might another assumption we are making differently. While yes libtock-c sort of has the expected nice round number where it implicitly expects apps to be placed, we don't explicitly have an intended load address. We link the binary several times with different fixed addresses in flash, let elf2tab add a reasonable header, and then expect the loader tool (in my case tockloader) to put the tbf in flash in a location where the kernel will run it.




> One could see this change as a UX adjustment

I think the way forward is to actually forget about libtock-c and libtock-rs and instead think about what we want the interface to elf2tab to be. I generally don't like adding features to elf2tab to support non-pic apps which are not a good fit for the tock linked-list-of-apps model. I also would like to see libtock-rs implement full TAB support (as libtock-c has where one compilation supports [nearly] every board, with multiple apps) before using it as a model for designing this interface.

One of the aspects that I _do_ like about `--protected-region-size` is it is not non-pic specific. In fact, one could argue the feature has nothing to do with pic/non-pic, and that it is only there so that elf2tab can help create tbfs with additional space in the protected region for the kernel to use. This particularly relevant if also using elf2tab to sign tbfs.

The drawback to all command line flags in elf2tab is that they apply to all elfs, `--protected-region-size` included. What about allowing the protected region size to be set via an elf symbol? That is a feature that all elfs/tbfs could use if they were so inclined. In contrast, something like a `tbf_start_address` symbol has no meaning for a pic app.

How a userland compilation framework generates these symbols is up to it (and outside the scope of this PR).

> Also (at least in the current implementation) the headers shouldn't exceed 256 bytes.

We should fix this.

lschuermann:
> What about allowing the protected region size to be set via an elf symbol?

Ah! I mean, this is almost exactly what this change does (it just sets an absolute address instead, and calls it `tbf_header`). I'll happily refactor this to work with a `protected_region_size` symbol!

bradjc:
> > What about allowing the protected region size to be set via an elf symbol?
> 
> Ah! I mean, this is almost exactly what this change does (it just sets an absolute address instead, and calls it `tbf_header`). I'll happily refactor this to work with a `protected_region_size` symbol! And I do agree that this is more general and has better semantics over the `tbf_header` symbol.

We would need to establish what happens if both are set. Probably the elf overrides the command line.

lschuermann:
> We would need to establish what happens if both are set. Probably the elf overrides the command line.

I would've vouched for the other way around. My guess is that build infrastructures will adopt this option (at least `libtock-rs` in the short term, to actually build working binaries with recent elf2tab versions). The command line would give users a convenient way to override the toolchain's default value. I would print out a warning in case one overrides the other (the current patch would do so already).

But I don't have strong feelings either way. If you'd prefer the ELF symbol to take precedence, I'd be fine with that too.

lschuermann:
I have updated the code, commit message and pull request description according to this discussion. I'm happy to change the ELF-symbol to take precedence over the command line, if desired. Tested this to work with `libtock-rs` and `libtock-c` apps. Let me know what you think.

bradjc:
> But I don't have strong feelings either way. If you'd prefer the ELF symbol to take precedence, I'd be fine with that too.

My thinking being that if an elf goes to the trouble of specifying the protected size then it's probably important to use that. A user might want to set a default for all other TBFs via the command line.

bradjc (commented):
Looks good. Need to update the readme.

I think we shouldn't override the elf setting with the command line. It might not make sense to set all tbfs to the same protected size, but it might be useful to be able to set a default.

lschuermann:
Updated the readme and the ELF symbol / command line parameter precedence.

I do now realize why it would make sense for the symbol to take precedence: I missed the fact that elf2tab can consider multiple ELFs simultaneously and pack them into a TAB archive. I agree with the argument that it makes sense to add different protected region sizes for different TBFs in a single TAB.

bradjc (approved):
This works with the current libtock-c, correct?

Issue #77, opened 2023-08-08
I can't generate an RSA-signed tab - KeyRejected("InvalidEncoding")


george-cosma:
Hello! I was recently experimenting with elf2tab and I saw that I could sign an app using RSA4096. I tried following the tutorial on the README, but I can't get it working. I did this on the c_hello example from libtock-c.

Error:
```
thread 'main' panicked at 'RSA4096 could not be parsed: KeyRejected("InvalidEncoding")', /home/<user>/.cargo/registry/src/github.com-1ecc6299db9ec823/elf2tab-0.11.0/src/convert.rs:998:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: elf2tab::convert::elf_to_tbf
   3: elf2tab::main
```
Commands executed:
```bash
$ openssl genrsa -aes256 -out tockkey.private.pem 4096
$ openssl rsa -in tockkey.private.pem -outform der -out tockkey.private.der
$ openssl rsa -in tockkey.private.pem -outform der -pubout -out tockkey.public.der
```

```bash
$ openssl version
OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)
```

```bash
elf2tab --rsa4096-private tockkey.private.der --rsa4096-public tockkey.public.der -n c_hello --stack 2048 --app-heap 1024 --kernel-heap 1024 --kernel-major 2 --kernel-minor 0 -v -o build/c_hello.tab build/cortex-m0/cortex-m0.elf build/cortex-m3/cortex-m3.elf build/cortex-m4/cortex-m4.elf build/cortex-m7/cortex-m7.elf
# basically what `make` executes inside libtock-c, just with the rsa parameters added.
```

Just for context, here's where I think the issue is occuring: [let key_pair = signature::RsaKeyPair::from_der(&private_key_der)](https://github.com/tock/elf2tab/blob/bf5f46071e56642cd8d73e2bc84f5bc2582a212b/src/convert.rs#L996-L1000)

bradjc:
Interesting. I just copied these steps and they worked for me.

```
[...]

PR #78, opened 2023-08-08
Use privatekey.pk8 instead for rsa signatures

bradjc:
It seems that different OpenSSL versions create different .der types (#77), and the ring library we are using doesn't necessarily support them all. Try using .pk8 instead.

While we are at it, we do not need a separate library for public keys and we can get the public key from the private key pk8 file.


bradjc:
@phil-levis

Issue #94, opened 2025-05-28
Relocation data gets duplicated in the resulting TBF file


pekangas:
The last of the issues we found while catching up with top-of-tree elf2tab...

The logic that gathers relocation data and places it at the end of the file ends up also emitting the data in place, which results in relocations being in the file twice.

My fix is a bit ugly, basically skipping segments that only have relocation data in them. I believe it should generally work, unless some configurations end up combining relocations with other data in the same segment?

PR #95, opened 2025-05-28
Do not duplicate relocation data

pekangas:
Skip relocation data while processing segments since it will be added separately at the end of the file. Note that this logic skips entire segments if they contain nothing but relocation data so may break if a segment contains both relocations an other sections (but is that possible?)

Also handle .rela relocation sections

This fixes https://github.com/tock/elf2tab/issues/94 at least for our use cases..

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -663,37 +663,22 @@ pub fn elf_to_tbf(
         }
 
         // Insert padding between segments if needed.
-        if let Some(last_segment_address_end) = last_segment_address_end {
+        let padding = if let Some(last_segment_address_end) = last_segment_address_end {
             // We have a previous segment. Now, check if there is any padding
             // between the segments in the .elf.
             let chk_padding = (segment.p_paddr as usize).checked_sub(last_segment_address_end);
-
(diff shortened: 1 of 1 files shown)


src/convert.rs
```diff
@@ -814,12 +814,39 @@ pub fn elf_to_tbf(
             }
         }
 
-        // Save the end of this segment so we can check if padding is required
-        // between segments.
-        last_segment_address_end = Some(end_segment as usize);
+        if reloc_only {
+            if verbose {
+                println!("    Skipping segment since it only contains relocs");
+            }
```
bradjc, line 820: Can we just put a `continue` here (consistent with the checks above around line 629?

src/convert.rs
```diff
@@ -814,12 +814,39 @@ pub fn elf_to_tbf(
...
+        } else {
+            // Handle padding now that we know we want something from this
```
bradjc, line 822: Can we just move the entire padding check from above here? I'd like to keep the utility of the option of `last_segment_address_end` and not have `padding == 0` mean both no padding and there can't be padding because this is the first segment.

PR #96, opened 2025-08-05
Update to rust-toolchain toml file

brghena:
Update the rust-toolchain file to the rust-toolchain.toml version.

Apparently the old version is deprecated. Tock switched over in 2023 via https://github.com/tock/tock/pull/3593

I copied the components list from Tock. Not sure what's needed for elf2tab, but this seemed reasonable.

I tested this by running `cargo build` and it still worked.


brghena:
Attempting to revise CI, as the CI tools we _were_ using are all deprecated.

brghena:
I guess that worked. I think this is good to go.

bradjc (approved):

PR #102, opened 2025-11-25
Added total size padding override. Documented unused parameter

### LINKED ISSUE #100: Allow setting non-default padding
Currently the padding is set solely based on the architecture assumptions:
https://github.com/tock/elf2tab/blob/06b05fdad764c59c279a50686ccd512f544fb824/src/convert.rs#L192-L204

However these assumptions are not always correct. For instance, in case of RISC-V, the MPU (PMP) might have greater alignment requirements (see https://github.com/tock/tock/issues/4417). This can be *kind of* worked around by adding something like

```ld
    .dummy : {
        . = ALIGN(4096); /* For 4096 granularity */
    } > FLASH
```

but it also breaks as `elf2tbf` adds at least 4 bytes for relocation data to the total size.

### LINKED ISSUE #101: `--minimum-ram-size` option is not respected
The `--minimum-ram-size` option seem to not do anything. It is associated with the `minimum_stack_size` variable for some reason, which is not being used further in the code.
https://github.com/tock/elf2tab/blob/06b05fdad764c59c279a50686ccd512f544fb824/src/cmdline.rs#L64-L72

ZhekaS:
Fixes https://github.com/tock/elf2tab/issues/101
Fixes https://github.com/tock/elf2tab/issues/100

~~- Added an option `--total-size-padding <padding>` to override the default set by the specific architecture~~
- Added an optional ELF symbol `tbf_total_size_padding` lookup that sets the override of the architecture specific padding
- Added a note about `--minimum-ram-size` that it has no effect

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ Options:
       --disable                                                Mark the app as disabled in the TBF flags
       --app-version <APP_VERSION>                              Set the version number [default: 0]
       --minimum-ram-size <min-ram-size>                        in bytes - has no effect, kept for backwards compatibility
-      --total-size-padding <padding>                           pad the total size of the binary to the multiple of <padding>. If not specified, use the architecture default
   -o, --output-file <filename>                                 output file name [default: TockApp.tab]
   -n, --package-name <pkg-name>                                package name
       --stack <stack-size>                                     in bytes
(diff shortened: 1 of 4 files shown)


bradjc:
>     * Added a note about `--minimum-ram-size` that it has no effect

Yes.



>     * Added an option `--total-size-padding <padding>` to override the default set by the specific architecture

I have two reservations about this option.

1. It applies to all architectures. This would easily make invalid arm apps, or very large riscv apps.
2. I think I generally understand the issue this is trying to address, but this option could hide an annoying failure case for users: they have two TABs for the same app, with the same architecture list, and one works and one doesn't. We've skirted around this issue by maintaining the assumption that architecture == MPU requirements.

    If we want to change that assumption, I would prefer that it be explicit so that users and tockloader can reason about the correct tbf to install, and when a tbf will not work on their board.

    We could have tockloader handle this. Or, we could include all riscv apps twice at different paddings. Or we could just change elf2tab to use the larger padding. Or we could have a more targeted flag to build with a different size. But I would like to have the upstream elf2tab continue to work across tock boards.

ZhekaS:
> 2. I think I generally understand the issue this is trying to address, but this option could hide an annoying failure case for users: they have two TABs for the same app, with the same architecture list, and one works and one doesn't. We've skirted around this issue by maintaining the assumption that architecture == MPU requirements.
>    If we want to change that assumption, I would prefer that it be explicit so that users and tockloader can reason about the correct tbf to install, and when a tbf will not work on their board.
>    We could have tockloader handle this. Or, we could include all riscv apps twice at different paddings. Or we could just change elf2tab to use the larger padding. Or we could have a more targeted flag to build with a different size. But I would like to have the upstream elf2tab continue to work across tock boards.

OK, I think I see the issue that I totally missed when working on this PR, that the TAB might actually contain many TBFs with different size requirements. I guess a more targeted flag is indeed required.

EDIT:
@bradjc Do you think something like `--total-size-padding=<arch>,<padding>` would work?

EDIT2:
Alternatively, limit this functionality to the case when only a single ELF is supplied (however this is more of a workaround for a very specific case, as we might want to pack multiple ELFs for the same arch).

ZhekaS:
@bradjc I think the most "granular" control could be achieved by optionally passing the required padding size via a metadata in the ELF file itself, similarly to `tbf_protected_region_size`. ~~Should I update the PR with the proposal?~~
Updated the PR with the proposal

